PR #146, opened 2026-01-29
feat(HostDetailsPanel): replace Input with Textarea for startup command

RiceWays:
- Replace Input component with Textarea for startup command field
- Update className to use min-h-[80px] with font-mono and text-sm styling
- Add rows={3} prop to Textarea for better multi-line command support
- Import Textarea component from ui/textarea module
- Improve UX for entering longer or multi-line startup commands


RiceWays:
@codex review

PR #190, opened 2026-02-07
fix: restore built-in text editor paste behavior

binaricat:
## Summary
- fix global hotkey interception so Monaco editor areas are treated as editable and SFTP shortcuts are not consumed at app level
- allow native context menu for editable/Monaco regions from root container context-menu guard
- add Electron clipboard fallback bridge (`netcatty:clipboard:readText`) and expose it in preload/types
- add `useClipboardBackend` in application state and wire editor `Cmd/Ctrl+V` fallback to explicit clipboard read + insert
- keep Monaco context menu disabled in editor so right-click uses native OS paste path in Electron

## Validation
- `npx eslint App.tsx components/TextEditorModal.tsx components/sftp/hooks/useSftpKeyboardShortcuts.ts components/sftp-modal/hooks/useSftpModalKeyboardShortcuts.ts application/state/useClipboardBackend.ts`

## User-visible result
- built-in text editor supports paste reliably via keyboard and right-click paste in Electron

### DIFF
--- a/components/TextEditorModal.tsx
+++ b/components/TextEditorModal.tsx
@@ -269,11 +269,16 @@ export const TextEditorModal: React.FC<TextEditorModalProps> = ({
     const selections = editor.getSelections();
     if (!selections || selections.length === 0) return;
 
+    // Match Monaco's default multicursorPaste:'spread' behavior:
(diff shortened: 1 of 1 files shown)


binaricat:
@codex review

PR #209, opened 2026-02-14
security: harden external navigation / window.open

RoryChou-flux:
Fixes a security issue where `setWindowOpenHandler` allowed arbitrary http(s) URLs to open in an in-app window.

## Changes
- Deny-by-default for `window.open` / `setWindowOpenHandler`.
- Allowlist known OAuth hosts for in-app auth windows (with `sandbox: true`).
- Everything else opens via `shell.openExternal` to reduce remote-content & phishing surface.
- Ensure renderer `window.open` uses `noopener,noreferrer` and blocks non-http(s).
- Limit preload `window.netcatty` exposure to trusted origins only.

## How to test
1. Trigger an OAuth flow or external link.
2. Confirm OAuth opens in-app; other links open in the system browser.
3. Confirm popups cannot open arbitrary remote pages inside the app.

### DIFF
--- a/electron/preload.cjs
+++ b/electron/preload.cjs
@@ -858,7 +858,14 @@ function getAllowedRendererOrigins() {
       const u = new URL(devServerUrl);
       origins.add(u.origin);
       // Vite often binds to 0.0.0.0, but Chromium navigates via localhost.
-      if (u.hostname === "0.0.0.0" || u.hostname === "127.0.0.1" || u.hostname === "::1") {
+      if (
+        u.hostname === "0.0.0.0" ||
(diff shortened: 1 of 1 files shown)


RoryChou-flux:
Addressed: preload dev-origin allowlist now matches windowManager normalization for IPv6 forms (adds `[::1]`, `::`, `[::]` etc) so `window.netcatty` is exposed correctly when `VITE_DEV_SERVER_URL` uses bracketed IPv6.

Commit: c793aad8de3bd8a82a24828feb80cf7abca89ee9

PR #211, opened 2026-02-24
feat: tab bar orientation switching and drag-to-resize

RoryChou-flux:
## Summary
- Support switching the tab bar between **horizontal (top)** and **vertical (sidebar)** modes via right-click context menu or Settings > Appearance
- Vertical sidebar width is drag-resizable (140–400px), horizontal bar height is fixed at 32px
- Extract `WindowControls` into standalone component; add `ThinTitlebar` for vertical mode

## Details
- **Performance**: Replaced heavy Radix `ContextMenu` wrapper (portal + animation + focus trap) with lightweight custom positioned div for orientation toggle. Memo comparator only checks visual data props to prevent cascade re-renders from unstable callback references (`onSyncNow`, `onOpenSettings`)
- **Resize**: Uses direct DOM mutation during drag (bypass React reconciliation), commits to state only on mouseup
- **Persistence**: Orientation and size saved to localStorage with cross-window sync via IPC
- **i18n**: Full English and Chinese translations

## Test plan
- [ ] Default horizontal mode works identically to before
- [ ] Right-click tab bar blank area → context menu to switch orientation
- [ ] Vertical sidebar: drag right edge to resize (140–400px), persists after refresh
- [ ] Vertical mode: tab drag-reorder works vertically, scroll + fade masks
- [ ] Settings > Appearance > Tab Bar Position selector works
- [ ] macOS: traffic lights in ThinTitlebar; Windows/Linux: window controls in ThinTitlebar
- [ ] Cross-window sync when orientation changes

### DIFF
--- a/App.tsx
+++ b/App.tsx
@@ -309,9 +309,13 @@ function App({ settings }: { settings: SettingsState }) {
     onUpdateManagedSources: updateManagedSources,
   });
 
+  // Stable ref wrapper: keeps identity constant so TopTabs memo comparator
+  // can safely exclude onSyncNow without risking stale closures.
+  const syncNowRef = useRef(handleSyncNow);
+  syncNowRef.current = handleSyncNow;
   const handleSyncNowManual = useCallback(() => {
-    return handleSyncNow({ trigger: 'manual' });
-  }, [handleSyncNow]);
+    return syncNowRef.current({ trigger: 'manual' });
+  }, []);
 
   // Update check hook - checks for new versions on startup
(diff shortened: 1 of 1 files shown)


binaricat:
Thanks again for putting this PR together and for the effort on this feature.

After reviewing and trying it locally, we found that the current behavior still differs from the expected UX on our side. To avoid merging something that doesn't fully match the target experience, we'll close this PR for now.

If you're open to it, please feel free to submit a follow-up PR aligned with the expected interaction and visual behavior, and we'd be happy to review again.

PR #214, opened 2026-02-25
fix(auth): auto-submit filled keyboard-interactive prompts

RoryChou-flux:
## Summary
- Auto-fill password-like keyboard-interactive prompts from saved password.
- Auto-submit keyboard-interactive requests when all prompts are already filled (prevents modal flash).
- Prefer keyboard-interactive before password in password-only auth flows to avoid duplicate OTP challenges.

## Behavior Notes
- No behavior change for key-only or agent-first flows.
- Keyboard-interactive remains a fallback for non-password-only paths.

## Validation
- `npm run -s lint -- components/KeyboardInteractiveModal.tsx`
- `node --check electron/bridges/sshAuthHelper.cjs`

### DIFF
--- a/electron/bridges/sshAuthHelper.cjs
+++ b/electron/bridges/sshAuthHelper.cjs
@@ -329,47 +329,50 @@ async function findAllDefaultPrivateKeys(options = {}) {
       methods: authMethods.map(m => m.id),
     });
   
-   // Use dynamic authHandler to try all keys
-   let authIndex = 0;
-   const attemptedMethodIds = new Set();
-   
-   const authHandler = (methodsLeft, partialSuccess, callback) => {
-     const availableMethods = methodsLeft || ["publickey", "password", "keyboard-interactive", "agent"];
-     
-     while (authIndex < authMethods.length) {
-       const method = authMethods[authIndex];
-       authIndex++;
-       
-       if (attemptedMethodIds.has(method.id)) continue;
-       attemptedMethodIds.add(method.id);
-       
+  // Use dynamic authHandler to try all configured methods.
+  // Important: only mark a method as attempted when it is actually sent to ssh2.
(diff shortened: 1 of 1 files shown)


binaricat:
Thanks for the work here. I see the intent, but I think the current password detection approach has logic-level issues and is risky to ship as-is.

The core problem is that the implementation infers "password prompt" from free-form prompt text. In keyboard-interactive auth, prompt text is server-defined and highly variable, so text matching is inherently heuristic and can be wrong.

Specific issues:
1. False negatives/false positives from keyword matching
- Prompt wording varies widely by server/vendor/language; fixed regex terms will miss valid password prompts.
- Broad keywords can also match non-password challenges.

2. `echo=false` is not a reliable password signal
- Non-password sensitive fields can also be hidden (`echo=false`), so this check is necessary but not sufficient.

3. Multi-step MFA flows are not structurally distinguishable by text alone
- Keyboard-interactive can include password + OTP + security questions in arbitrary order and wording.
- Text heuristics cannot robustly separate these categories across environments.

4. Security/privacy risk of auto-fill + auto-submit
- If classification is wrong, saved password can be sent to an unintended prompt.
- Auto-submit removes user verification and amplifies the impact of misclassification.

5. Localization and custom prompts make behavior unstable
- Current patterns only cover a subset of languages and terminology.
- Enterprise/bastion systems often use custom prompt labels.

Given these issues, I think this should not rely on prompt text alone for automatic submission. A safer baseline would be:
- prefill only (no auto-submit) on first-seen challenge,
- challenge fingerprinting per host (name/instruction/prompts) before enabling trusted auto-submit,
- explicit user opt-in for keyboard-interactive auto-submit.

Because of the above logic and safety concerns, I’m closing this PR for now.

PR #217, opened 2026-02-28
fix(terminal): initialize xterm WebGL addon with valid constructor args

RoryChou-flux:
## Summary
- remove invalid object argument passed to `new WebglAddon(...)`
- use the supported constructor signature directly
- add a short code comment explaining why

## Why
Passing an object here is interpreted as a truthy `preserveDrawingBuffer` value, which can cause intermittent rendering artifacts/ghosting in terminal output (especially with CJK text and heavy output).

## Scope
- only touches terminal runtime WebGL addon initialization
- no behavior change outside renderer setup


binaricat:
Thanks for the contribution, @RoryChou-flux! 🎉 Clean and correct fix. 
Merging now!

PR #235, opened 2026-03-03
fix(sftp): prevent stale session race when reopening modal

binaricat:
## Summary
- prevent stale SFTP session ID reuse when modal closes and reopens quickly
- serialize close/open with an in-flight close promise gate
- keep hide-with-active-transfer behavior while restoring awaited close path when no active task

## Changes
- clear session id ref before awaiting backend close
- make ensureSftp wait for pending close completion
- route reconnect close through closeSftpSession
- close immediately in handleClose when no active transfers, defer close after hide only when transfers are active

## Validation
- npx eslint components/SFTPModal.tsx components/sftp-modal/hooks/useSftpModalSession.ts

### DIFF
--- a/components/SFTPModal.tsx
+++ b/components/SFTPModal.tsx
@@ -205,6 +205,7 @@ const SFTPModal: React.FC<SFTPModalProps> = ({
     loading,
     setLoading,
     reconnecting,
+    sessionVersion,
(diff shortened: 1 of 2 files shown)


binaricat:
@codex review

PR #295, opened 2026-03-09
fix: disable context menu in alternate screen to prevent tmux double menu

binaricat:
## Summary

Fix duplicate context menus when using tmux with mouse mode enabled (`set -g mouse on`).

### Problem

When terminal applications enable **mouse tracking** (e.g., tmux `set -g mouse on`, vim `set mouse=a`), right-clicking shows **two context menus** simultaneously:
1. The terminal application's own menu (e.g., tmux's context menu)
2. Netcatty's built-in context menu

### Solution

Detect when the terminal application has enabled **mouse tracking mode** via xterm.js `term.modes.mouseTrackingMode`, and disable Netcatty's context menu in that state. This lets the terminal application handle mouse events natively without interference.

This approach is more precise than checking for alternate screen — it only disables the context menu when the application is **actively capturing mouse events**:

| Scenario | Mouse Tracking | Context Menu |
|---|:---:|:---:|
| Normal terminal | ❌ | Netcatty menu ✅ |
| tmux + `set -g mouse on` | ✅ | Passthrough to tmux ✅ |
| tmux without mouse mode | ❌ | Netcatty menu ✅ |
| vim + `set mouse=a` | ✅ | Passthrough to vim ✅ |
| vim without mouse | ❌ | Netcatty menu ✅ |
| less / man | ❌ | Netcatty menu ✅ |

### Changes

- **`Terminal.tsx`**: Added `hasMouseTracking` state tracking via `onWriteParsed` + `term.modes.mouseTrackingMode`, with ref-guard to avoid unnecessary re-renders during heavy terminal output
- **`TerminalContextMenu.tsx`**: Added `isAlternateScreen` prop; when true, disables the Radix context menu and bypasses custom right-click handlers (paste/select-word)

Partially addresses #294

### DIFF
--- a/components/Terminal.tsx
+++ b/components/Terminal.tsx
@@ -215,7 +215,8 @@ const TerminalComponent: React.FC<TerminalProps> = ({
   const hasConnectedRef = useRef(false);
   const hasRunStartupCommandRef = useRef(false);
   const commandBufferRef = useRef<string>("");
-  const [isAlternateScreen, setIsAlternateScreen] = useState(false);
+  const [hasMouseTracking, setHasMouseTracking] = useState(false);
+  const mouseTrackingRef = useRef(false);
   const serialLineBufferRef = useRef<string>("");
 
   const terminalSettingsRef = useRef(terminalSettings);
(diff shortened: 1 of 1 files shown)


binaricat:
@codex review

PR #301, opened 2026-03-09
feat: support system theme auto-switching

binaricat:
## Summary

Support automatic UI theme switching based on OS system preference.

### Problem

Previously, users could only manually switch between Light and Dark themes. There was no way to follow the system's color scheme preference automatically.

### Solution

Add `system` as a third theme option. When selected, the UI automatically follows the OS color scheme and switches in real-time when the system appearance changes (e.g., macOS Automatic appearance).

### Changes

| File | Change |
|------|--------|
| `useSettingsState.ts` | Add `resolvedTheme` state derived from `matchMedia('prefers-color-scheme: dark')`. Add `change` event listener for real-time system preference tracking. Update `applyThemeTokens` to use `resolvedTheme` |
| `SettingsAppearanceTab.tsx` | Replace dark mode Toggle with 3-segment selector (Light / System / Dark) using Sun/Monitor/Moon icons |
| `en.ts` / `zh-CN.ts` | Replace `darkMode` i18n keys with new `theme` keys including `system` option |

### Design

- `theme` (persisted): `'light' | 'dark' | 'system'` — the user's preference
- `resolvedTheme` (derived): always `'light' | 'dark'` — the effective theme used for rendering
- Default changed from `'light'` to `'system'` for new users
- Backward compatible: existing users with `'light'` or `'dark'` stored will keep their choice

Partially addresses #294

### DIFF
--- a/App.tsx
+++ b/App.tsx
@@ -167,8 +167,8 @@ function App({ settings }: { settings: SettingsState }) {
   const [passphraseQueue, setPassphraseQueue] = useState<PassphraseRequest[]>([]);
 
   const {
-    theme,
     setTheme,
+    resolvedTheme,
     setTerminalThemeId,
     currentTerminalTheme,
     terminalFontFamilyId,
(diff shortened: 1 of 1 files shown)


binaricat:
@codex review

PR #305, opened 2026-03-10
fix(sftp): handle non-fatal agent auth errors for MFA/keyboard-interactive (#304)

### LINKED ISSUE #304: bug:sftp 连接存在问题
连接ESXI sftp报错，怎么输入密码都不让连接，只有SSH正常：

Image

Image

同类产品WT没问题

Image

binaricat:
## Problem

Closes #304

ESXi (and other servers requiring keyboard-interactive/MFA authentication) fail to connect via SFTP with `"Error: Failed to connect to agent"`, while SSH works correctly.

## Root Cause

Two compounding bugs:

### 1. `ssh2-sftp-client` error listener kills auth fallback

When ssh2 tries agent auth and the agent is unavailable, it emits an `error` event with `err.level = 'agent'`, then internally calls `tryNextAuth()` to fall through to keyboard-interactive. However, `ssh2-sftp-client`'s `connect()` installs temporary error listeners that immediately **reject the entire connection promise** on any error — preventing the auth flow from reaching keyboard-interactive.

SSH was unaffected because it uses raw `ssh2.Client` directly.

### 2. Windows SSH Agent service not checked  

`getSshAgentSocket()` on Windows unconditionally returned `\\.\pipe\openssh-ssh-agent` without checking if the SSH Agent service is actually running. This caused agent auth to always be attempted on Windows when the service was stopped — a very common scenario.

## Fix

1. **Bypass `ssh2-sftp-client.connect()`** — use direct `ssh2.Client.connect()` with manual SFTP channel opening. Filter `err.level === 'agent'` errors as non-fatal.
2. **Add `getAvailableAgentSocket()`** — async validation that checks `sc query ssh-agent` on Windows before returning the agent pipe path.
3. **Jump host chain** — same error filtering applied.

## Testing

- Requires a server with keyboard-interactive (MFA) auth to verify end-to-end
- Workaround posted in #304 for user to verify the root cause

### DIFF
--- a/electron/bridges/sftpBridge.cjs
+++ b/electron/bridges/sftpBridge.cjs
@@ -946,6 +946,14 @@ async function openSftp(event, options) {
     const sshClient = client.client;
 
     await new Promise((resolve, reject) => {
+      let settled = false;
+      const settle = (fn, val) => {
+        if (settled) return;
+        settled = true;
+        cleanup();
+        fn(val);
+      };
+
       const onError = (err) => {
         // Filter out non-fatal authentication errors.
         // ssh2 sets err.level = 'agent' when agent auth fails — it then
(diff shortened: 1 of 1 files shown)


binaricat:
@codex review

PR #311, opened 2026-03-11
fix(sftp): drag-upload goes to wrong directory after navigation

binaricat:
## Summary

- Fix drag-and-drop upload targeting the previous directory after navigating to a new one
- Update `currentPath` immediately when navigation begins, not after async file listing completes

## Problem

When dragging external files to upload via SFTP:
1. In remote directory A, drag upload → works
2. Navigate to directory B, drag upload → file uploads to directory A (wrong!)
3. Drag upload again → correctly uploads to directory B

**Root cause**: In `navigateTo()`, the cache-miss path sets `loading: true` without updating `currentPath`. The path was only updated after the async file listing completed. During this window, `getActivePane()` returns the pane with the stale `currentPath`, so uploads target the old directory.

## Fix

Update `currentPath` immediately when setting `loading: true` (line 121 of `useSftpPaneActions.ts`), matching the behavior of the cache-hit path which already updates `currentPath` synchronously. The subsequent update after file listing is now idempotent.

## Test plan

- [ ] Connect to SFTP server, navigate to directory A
- [ ] Drag a file from local to upload → verify it goes to A
- [ ] Navigate to directory B, immediately drag a file to upload → verify it goes to B (not A)
- [ ] Repeat with cache-hit navigation (navigate back to A within 10s) → verify upload still targets correct directory

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -7,7 +7,7 @@ import reactHooks from "eslint-plugin-react-hooks";
 export default [
   js.configs.recommended,
   {
-    ignores: ["node_modules/**", "dist/**", "electron/**", "scripts/**", "public/monaco/**", ".github/**"],
+    ignores: ["node_modules/**", "dist/**", "electron/**", "scripts/**", "public/monaco/**", ".github/**", ".claude/**"],
   },
   {
     files: ["**/*.{ts,tsx}"],


binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

PR #367, opened 2026-03-17
fix: open settings window on the same display as main window

binaricat:
## Summary

- Center settings window on the same display as the main window using `screen.getDisplayMatching()`
- Fixes issue where settings opens on the laptop screen when the main window is on an external monitor

Ref #294

## Test plan

- [ ] Move main window to external display, open Settings → should appear on external display
- [ ] Single display setup still works as before
- [ ] Main window on primary display → Settings appears on primary display


binaricat:
@codex review

PR #413, opened 2026-03-20
Add multi-host AI terminal execution visibility and context management

Zhangtiande:
## Summary

- Add multi-connection AI terminal context, allowing the AI to target specific terminal sessions when multiple connections are open
- Show AI execution target hosts in the sidebar and always display AI context hosts for better visibility
- Show terminal commands in tool call titles and make terminal execute targets visible in chat
- Add dedicated `TerminalToolCall` component with hash-based per-host accent colors and a two-row layout (host name + command) with colored left border for visual identity
- When no host info is available (e.g. ACP/external agents), display the command as primary content in a single-row layout instead of a redundant "terminal_execute" label
- Fix `areMessagesEqual` to compare `attachments`/`images` fields, preventing stale renders when user messages with attachments are updated
- Fix AI extra session picker toggling behavior

## Test plan

- [x] Open multiple terminal connections and verify the AI sidebar shows all context hosts
- [x] Send an AI command targeting a specific host and verify the correct host is highlighted in chat
- [x] Verify `TerminalToolCall` renders with distinct accent colors per host
- [x] Verify ACP/external agent terminal calls display the command directly without a host label
- [x] Verify messages with attachments re-render correctly after updates

### DIFF
--- a/components/ai-elements/terminal-tool-call.tsx
+++ b/components/ai-elements/terminal-tool-call.tsx
@@ -1,0 +1,152 @@
+import { CheckCircle2, ChevronDown, ChevronRight, Loader2, Slash, XCircle } from 'lucide-react';
+import { useState } from 'react';
+import { cn } from '../../lib/utils';
+
+// Palette of accent colors for host identity (works on light and dark backgrounds)
+const HOST_COLORS = [
+  '#60a5fa', // blue
+  '#a78bfa', // violet
+  '#34d399', // emerald
+  '#fb923c', // orange
+  '#f472b6', // pink
+  '#38bdf8', // sky
+  '#facc15', // yellow
+  '#f87171', // red
+];
+
+function getHostColor(sessionId: string): string {
+  let h = 0;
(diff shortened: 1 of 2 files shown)


binaricat:
Thanks a lot for the thoughtful work here. I actually considered a similar direction before, and I think the implementation is well thought out.

That said, after sitting with the interaction model a bit more, I still feel the AI sidebar/context for each tab should stay scoped to that tab itself. In my view, if a user wants to operate across multiple terminals, the right unit for that should be a workspace, rather than expanding a single tab’s context beyond itself.

I do think the broader idea is interesting though. A cross-host orchestration flow probably makes more sense as a future global AI page, where the AI can work across all hosts intentionally, instead of stretching the per-tab experience.

So I’m going to close this PR for now, not because the approach isn’t valuable, but because I think it’s heading in a different product direction from where I want Netcatty to go at the moment.

Really appreciate the time and care you put into this. I’d be very happy to revisit some of these ideas later if/when we design that global AI workflow.

PR #564, opened 2026-03-29
perf: optimize same-host SFTP transfer with remote cp command

### LINKED ISSUE #561: [Enhancement] sFTP: Don't download and reupload files on the same server
First of all: Really good job with the last update :3

I just noticed that opening the same host on both panels does not allow for seamless file moving or copying.

Netcatty always downloads the file and reuploads it afterwards.

I don't think this should be how it works, since it's the same host...

binaricat:
## Summary

Closes #561

When both SFTP panels are connected to the **same remote host**, file transfers currently download to a local temp file then re-upload — consuming 2x bandwidth and adding unnecessary latency.

This PR detects same-host transfers by comparing `hostId` on both sides, then uses SSH `exec('cp -a ...')` directly on the remote server instead of the download+upload round-trip.

## Changes

- **`transferBridge.cjs`** — New `sameHost` fast path before the existing `sftp→sftp` branch, using `execSshCommand` to run `cp -a` on the remote host
- **`useSftpTransfers.ts`** — Detect same-host by comparing `hostId` from both panes; propagate `sameHost` flag through `transferFile` and `transferDirectory`
- **`sftpBridge.cjs`** — Export existing `execSshCommand` for use by transferBridge
- **`global.d.ts`** — Add `sameHost` option to the transfer IPC type

## Behavior

- **Same host**: `cp -a` via SSH exec (instant, no bandwidth)
- **Different hosts**: Unchanged — still uses the existing download+upload flow
- **Fallback**: If SSH exec is unavailable, throws an error (conservative — won't silently degrade)

## Test plan

- [ ] Connect both panels to the same remote host, copy files between directories — verify no temp file download occurs
- [ ] Copy directories between panels on the same host — verify recursive copy works
- [ ] Connect panels to different hosts, transfer files — verify existing behavior unchanged
- [ ] Test with paths containing single quotes and special characters

### DIFF
--- a/electron/bridges/sftpBridge.cjs
+++ b/electron/bridges/sftpBridge.cjs
@@ -1858,5 +1858,4 @@ module.exports = {
   renameSftp,
   statSftp,
   chmodSftp,
-  execSshCommand,
 };
--- a/electron/bridges/transferBridge.cjs
+++ b/electron/bridges/transferBridge.cjs
@@ -6,7 +6,7 @@
 const fs = require("node:fs");
 const path = require("node:path");
 const os = require("node:os");
-const { encodePathForSession, ensureRemoteDirForSession, requireSftpChannel, execSshCommand } = require("./sftpBridge.cjs");
+const { encodePathForSession, ensureRemoteDirForSession, requireSftpChannel } = require("./sftpBridge.cjs");
 
 /**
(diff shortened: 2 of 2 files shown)


binaricat:
@codex review

binaricat:
@codex review

PR #570, opened 2026-03-29
fix: isolate SFTP actions and selection state across panes and tabs

### LINKED ISSUE #569: sFTP, actions are repeated for every single connection in active panel
I've noticed that actions like deleting a file are repeated across all connections of a panel.
See video for more information:

https://github.com/user-attachments/assets/9bd58630-dfc0-4431-9959-3e9b711682b9

This could lead to accidental data loss

binaricat:
## Summary
- fix SFTP keyboard-triggered dialogs so rename, delete, and new-folder actions only open in the currently visible SFTP surface
- stop selection state from leaking across panes and tabs by clearing the other pane only when the user actually selects a file or folder
- preserve the focused pane or tab correctly when switching sides, including the cross-side tab-click case where the newly selected tab should keep its own selection
- reset list and tree range-selection anchors after external clears so the next Shift selection starts fresh instead of reusing a stale range
- improve SFTP dialog context by showing the current host and path in destructive prompts

## Root cause
The original fix only scoped keyboard dialog handling to the active tab, but selection and focus state were still shared too broadly across mounted SFTP panes. That caused a few follow-up regressions:
- hidden SFTP surfaces could still react to dialog actions
- focus changes could clear the wrong pane or tab
- stale Shift-selection anchors could survive after cross-pane clears

This PR finishes that cleanup by making dialog routing target the active visible surface and tightening when and where selection cleanup happens.

Closes #569

## Test plan
- [x] `npm run lint`
- [x] `npm run build`
- [x] Open multiple SFTP tabs on one side and verify Delete / Rename / New Folder only affect the visible tab
- [x] Switch between left and right panes and confirm the first click keeps the new selection
- [x] Click directly into a non-active tab on the opposite side and confirm its existing selection is preserved
- [x] Clear selection via another pane, then use `Shift` selection in both list and tree views and confirm the range starts fresh

### DIFF
--- a/components/SftpView.tsx
+++ b/components/SftpView.tsx
@@ -40,7 +40,7 @@ import { useSftpViewPaneCallbacks } from "./sftp/hooks/useSftpViewPaneCallbacks"
 import { useSftpViewTabs } from "./sftp/hooks/useSftpViewTabs";
 import { useSftpKeyboardShortcuts } from "./sftp/hooks/useSftpKeyboardShortcuts";
 import { sftpFocusStore, SftpFocusedSide, useSftpFocusedSide } from "./sftp/hooks/useSftpFocusedPane";
-import { keepOnlyActivePaneSelections } from "./sftp/hooks/selectionScope";
+import { keepOnlyActivePaneSelections, keepOnlyPaneSelections } from "./sftp/hooks/selectionScope";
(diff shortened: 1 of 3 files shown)


binaricat:
@codex review

PR #599, opened 2026-04-01
Add Skills + CLI mode for external agents

tces1:
## Summary

This PR adds a new `Skills + CLI` integration mode for external ACP agents, while keeping the existing MCP path intact.

The main goal is to make external agents work reliably in environments where third-party MCP is unavailable or restricted, and to reduce overthinking / unnecessary discovery steps for simple Netcatty tasks.

### AI settings
- Added a new `Tool Access` setting to switch between:
  - `MCP`
  - `Skills + CLI`
- Persisted the selected mode in AI settings state/storage.
- Added localized UI copy, including a note that `Skills + CLI` currently requires a working `Node.js` runtime on the host.

### External agent flow
- Added a `Skills + CLI` path for ACP-backed external agents.
- In `Skills + CLI` mode, Netcatty no longer injects the Netcatty MCP server into the external agent.
- Instead, Netcatty provides:
  - a local skill file
  - an exact CLI invocation pattern
  - current chat/session context

### `netcatty-tool-cli`
- Added a new local CLI bridge:
  - `env`
  - `session`
  - `resource environment`
  - `exec`
  - `status`
  - `cancel`
  - `resume`
- Added a local RPC client and discovery path support for the CLI bridge.
- Reused existing main-process execution logic and safety checks rather than duplicating command execution behavior.

### Skills and prompt improvements
- Added a Netcatty skill for the CLI workflow.
- Split the skill into:
  - a short primary `SKILL.md`
  - focused reference docs for session types, control commands, and errors
- Tightened prompt guidance for simple read-only tasks:
  - prefer the shortest path
  - avoid extra script generation
  - avoid `$()` and backticks
  - avoid unnecessary `sh -c` / `bash -c` wrappers
- Prefer concise tabular output when appropriate.

### Session targeting optimization
- The host now provides the external agent with a default target session when it can determine one safely.
- For simple requests, the agent can use the default session directly instead of always rediscovering sessions first.
- Fallback discovery remains available when the target is ambiguous or the user explicitly references another session.

### Packaging
- Fixed packaged app behavior for `Skills + CLI` by ensuring:
  - skill files are available from unpacked app resources
  - CLI assets are available from unpacked app resources
  - packaged paths resolve correctly in `aiBridge`
- This avoids packaged-app failures where external agents could not find the Netcatty skill or CLI files.

### Reliability / cleanup
- Removed unused MCP preload helper code.
- Hardened CLI RPC behavior to avoid indefinite hangs on write/connection failures.
- Fixed execution lock handling for synchronous throws in command execution paths.

## Notes

- The MCP contract is intentionally left unchanged.
- The new CLI/session-targeting optimizations are scoped to the `Skills + CLI` path only.

## Testing

- `npm run build`
- `node --check electron/bridges/aiBridge.cjs`
- Manual validation of:
  - `MCP` vs `Skills + CLI` settings toggle
  - `Skills + CLI` execution against a live Netcatty session
  - packaged macOS app behavior from `/Applications/Netcatty.app`


tces1:
截屏2026-04-10 10 51 57

更新一下流程图
我本地用5.4 xhigh 跑了十几轮review，到这里基本没看到其他问题了

PR #666, opened 2026-04-09
fix: pin toolbar above content on KeychainManager page

yuzifu:
## Summary
- Restructure the KeychainManager layout so the toolbar (filter tabs, search, actions) stays pinned at the top while the list below scrolls independently
- Previously the toolbar scrolled away with the content when the key/snippet list was long, making filtering and searching inconvenient

## Changes
- Wrap the main content in a `flex flex-col min-h-0` container
- Add `shrink-0` to the toolbar so it never collapses
- Move `flex-1 overflow-y-auto` (and the slide-out panel offset `mr-[380px]`) onto a new inner scroll container that holds the keys section

## Test plan
- [x] Open the Keychain page with enough keys/snippets to require scrolling
- [x] Scroll the list — verify the toolbar stays visible at the top
- [x] Toggle a slide-out panel and verify the content area still reserves the 380px right margin
- [x] Verify filter tabs, search, and action buttons in the toolbar still work as before

### DIFF
--- a/components/KeychainManager.tsx
+++ b/components/KeychainManager.tsx
@@ -513,7 +513,12 @@ echo $3 >> "$FILE"`);
       />
 
       {/* Main Content */}
-      <div className="flex-1 flex flex-col min-h-0">
+      <div
+        className={cn(
+          "flex-1 flex flex-col min-h-0 transition-all duration-200",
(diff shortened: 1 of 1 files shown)


binaricat:
@codex review

PR #676, opened 2026-04-10
fix: prevent crash when clicking external links with no default browser

binaricat:
## Summary
- Prevent a main-process crash on systems (e.g. Tiny11) that have no default browser configured for http/https URLs
- Show a friendly toast instead of silently failing when `shell.openExternal()` fails

Closes #663

## Root cause

The user reported this crash log:
```
unhandledRejection
Failed to open: 没有应用程序与此操作的指定文件有关联。 (0x483)
Error: Failed to open: 没有应用程序与此操作的指定文件有关联。 (0x483)
```

Windows error `0x483` is `ERROR_NO_ASSOCIATION` — the OS has no handler registered for the URL scheme. Tiny11 users often have no default browser set, so `shell.openExternal()` rejects immediately.

Tracing the crash path:

1. Renderer clicks a link → calls `bridge.openExternal(url)` → IPC to main.
2. `main.cjs:656` handler `await shell.openExternal(url)` — rejects. Electron's `ipcMain.handle` forwards the rejection to the renderer as an IPC error (does not crash).
3. Renderer's `useApplicationBackend.openExternal` catches the IPC error and falls back to `window.open(url, "_blank", ...)`.
4. `window.open()` triggers main's `createExternalOnlyWindowOpenHandler` in `windowManager.cjs:381`, which does:
   ```js
   try {
     void shell?.openExternal?.(targetUrl);
   } catch { /* ignore */ }
   ```
5. **`void` does not catch Promise rejections.** It only discards the return value. `shell.openExternal` still returns a rejecting Promise that has no rejection handler — which becomes an `unhandledRejection`.
6. The global `unhandledRejection` handler in `main.cjs:78` re-throws the error to preserve fatal semantics, and the main process dies.

## Changes

- **`windowManager.cjs`** (the actual crash source): attach an explicit `.catch` on the openExternal Promise in both `createExternalOnlyWindowOpenHandler` and `createAppWindowOpenHandler` so rejections can never propagate. Kept the existing `try/catch` for synchronous errors, added the async catch as a sibling.
- **`main.cjs` IPC handler**: wrap `shell.openExternal` in try/catch and return a structured `{ success, error }` result. The renderer now gets a resolvable response on failure instead of an IPC error.
- **`global.d.ts`**: update the `openExternal` return type to `Promise<{ success: boolean; error?: string }>`.
- **`useApplicationBackend.ts`**: check the structured result and throw on failure so callers can present a user-facing message. Dropped the `window.open()` fallback from the Electron branch (it re-entered the buggy window-open handler and is now unreachable anyway). Kept the fallback only for non-Electron environments (tests, dev server).
- **`SettingsApplicationTab.tsx`**: wrap each of the 4 external-link action rows with a `handleOpenExternal` that shows a friendly toast ("No default browser configured — please set one in system settings") on failure.
- **`en.ts` + `zh-CN.ts`**: add `settings.application.openExternal.failedTitle` and `.failedBody` strings for the toast.

## Test plan

- [ ] Build and run on a Windows machine with a default browser set — verify all 4 links in Settings → Application still open in the browser as before.
- [ ] On a Tiny11 or similar stripped Windows without a default browser (or, for quick reproduction, temporarily unregister the default browser in Default Apps), click each link in Settings → Application:
  - [ ] The app does NOT crash
  - [ ] A toast appears with "Cannot open link — No default browser is configured…"
- [ ] Check the crash log UI — verify no `unhandledRejection` is logged for `openExternal` failures.
- [ ] Exercise other places that call `openExternal` (update check, cloud sync OAuth, AI settings, terminal link click) to make sure none regresses. Most of those already used their own `try/catch`.

## Notes

- The other callers of `openExternal` (`useUpdateCheck`, `useCloudSync`, `SettingsAITab`, `createXTermRuntime`) either already catch errors or use `void openExternal()`. They now receive a structured-result-aware rejection. Since they previously caught the old-style throw, they still behave the same (except they no longer silently open a blank `window.open()` window on failure).
- This PR only adds a toast in `SettingsApplicationTab` because that's where the issue was reported. Adding the same toast treatment to other call sites is a reasonable follow-up but out of scope here.

### DIFF
--- a/application/i18n/locales/en.ts
+++ b/application/i18n/locales/en.ts
@@ -198,7 +198,7 @@ const en: Messages = {
   'settings.application.whatsNew': "What's new",
   'settings.application.whatsNew.subtitle': 'Show release notes',
   'settings.application.openExternal.failedTitle': 'Cannot open link',
-  'settings.application.openExternal.failedBody': 'No default browser is configured for this URL scheme. Please set a default browser in your system settings and try again.',
+  'settings.application.openExternal.failedBody': 'The link could not be opened in either the system browser or the built-in browser window.',
   'settings.vault.title': 'Vault',
   'settings.vault.showRecentHosts': 'Show recently connected hosts',
   'settings.vault.showRecentHostsDesc': 'Display a section of recently connected hosts at the top of the vault',
--- a/application/i18n/locales/en.ts
+++ b/application/i18n/locales/en.ts
@@ -197,6 +197,8 @@ const en: Messages = {
   'settings.application.github.subtitle': 'Source code',
   'settings.application.whatsNew': "What's new",
   'settings.application.whatsNew.subtitle': 'Show release notes',
+  'settings.application.openExternal.failedTitle': 'Cannot open link',
+  'settings.application.openExternal.failedBody': 'No default browser is configured for this URL scheme. Please set a default browser in your system settings and try again.',
   'settings.vault.title': 'Vault',
   'settings.vault.showRecentHosts': 'Show recently connected hosts',
   'settings.vault.showRecentHostsDesc': 'Display a section of recently connected hosts at the top of the vault',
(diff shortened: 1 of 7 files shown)


binaricat:
@codex review

binaricat:
@codex review

PR #682, opened 2026-04-10
fix: pass legacyAlgorithms to port forwarding SSH connections

### LINKED ISSUE #678: Bug: [Feature Request] Add SSH algorithm compatibility options for legacy servers (MACs, KexAlgorithms)
## Problem Description

I'm trying to use netcatty to create an SSH local port forwarding through a jump server to access an internal server, but the connection fails when the SSH server uses older cryptographic algorithms.

## Environment

- **netcatty version**: [Please fill in your version]
- **OS**: Windows 10/11
- **Local SSH client**: OpenSSH_for_Windows_9.5p1
- **Remote SSH server**: OpenSSH_7.4

## Steps to Reproduce

1. Configure a local forwarding rule:
   - **Bind Address**: 127.0.0.1
   - **Local Port**: 10022
   - **Jump Server**: A server (32.110.38.175:22)
   - **Target Address**: 192.168.2.2:22
   
2. Click "Start" to initialize the tunnel

3. Error occurs immediately

## Error Messages
Error invoking remote method 'netcatty:start': Error: Connection lost before handshake

### netcatty GUI Error:

Image

binaricat:
## Summary
- Thread the host's `legacyAlgorithms` setting through to the port forwarding SSH connection, so tunnels to older SSH servers (e.g. OpenSSH 7.4) can negotiate legacy key exchange and cipher algorithms

Closes #678

## Root cause
The SSH terminal path correctly passes `legacyAlgorithms` through `buildAlgorithms()` to enable older algorithms like `diffie-hellman-group14-sha1`, `aes128-cbc`, and `ssh-rsa` when the host setting is toggled on. But the port forwarding path was never wired up — it always used modern-only algorithms, causing "Connection lost before handshake" when the jump server or target host only supports legacy algorithms.

## Changes (3 files, +6 / -1)
- **`sshBridge.cjs`**: export `buildAlgorithms` so `portForwardingBridge` can reuse it
- **`portForwardingBridge.cjs`**: destructure `legacyAlgorithms` from payload, add `algorithms: buildAlgorithms(legacyAlgorithms)` to `connectOpts`, thread `legacyAlgorithms` into the `connectThroughChain` options for jump host connections
- **`portForwardingService.ts`**: pass `host.legacyAlgorithms` in the `startPortForward` bridge call

## Test plan
- [ ] Configure a host with `legacyAlgorithms` enabled that runs an older SSH server (OpenSSH 7.4 or similar)
- [ ] Create a local port forwarding rule through that host
- [ ] Verify the tunnel connects successfully (no "Connection lost before handshake")
- [ ] Verify tunnels to modern SSH servers still work with `legacyAlgorithms` disabled
- [ ] Verify jump host chain forwarding also works with `legacyAlgorithms` enabled on the jump host

### DIFF
--- a/electron/bridges/portForwardingBridge.cjs
+++ b/electron/bridges/portForwardingBridge.cjs
@@ -10,7 +10,7 @@ const net = require("node:net");
 const { Client: SSHClient } = require("ssh2");
 const { NetcattyAgent } = require("./netcattyAgent.cjs");
 const keyboardInteractiveHandler = require("./keyboardInteractiveHandler.cjs");
-const { connectThroughChain } = require("./sshBridge.cjs");
+const { connectThroughChain, buildAlgorithms } = require("./sshBridge.cjs");
 const { createProxySocket } = require("./proxyUtils.cjs");
 const { 
   buildAuthHandler, 
@@ -59,6 +59,7 @@ async function startPortForward(event, payload) {
     proxy,
     jumpHosts = [],
     identityFilePaths,
+    legacyAlgorithms,
   } = payload;
 
   const conn = new SSHClient();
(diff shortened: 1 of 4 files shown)


binaricat:
@codex review

PR #698, opened 2026-04-12
[codex] Add vault root ungrouped host filter toggle

binaricat:
## What changed
- added a new setting in the Vault section of Appearance settings to let users show only ungrouped hosts in the root host list
- updated the root host list so that, when the toggle is enabled, it only shows hosts without a group while grouped hosts remain available through the sidebar groups
- wired the new setting into synced settings so the preference carries across devices
- added English and Chinese copy for the new setting

## Why this changed
Issue #638 asked for more control over what appears in the main host list. The current root view shows every host, which makes grouped hosts appear twice: once through their group and again in the root list. This toggle reduces clutter without changing grouped navigation.

## User impact
- users who prefer the current behavior do not need to change anything
- users who want a cleaner root view can enable the new toggle and keep the root list focused on ungrouped hosts only

## Validation
- `npm run lint`
- `npm run build`

### DIFF
--- a/App.tsx
+++ b/App.tsx
@@ -1469,6 +1469,8 @@ function App({ settings }: { settings: SettingsState }) {
             onClearUnsavedConnectionLogs={clearUnsavedConnectionLogs}
             onRunSnippet={runSnippet}
             onOpenLogView={openLogView}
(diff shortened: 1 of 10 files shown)


binaricat:
@codex review

PR #701, opened 2026-04-12
feat: preserve terminal buffer across reconnect (#695)

### LINKED ISSUE #695: 连接断开，重连之后，能否保持断开之前的屏幕记录，不清屏。
连接断开，重连之后，能否保持断开之前的屏幕记录，不清屏。

binaricat:
## Summary
- `handleRetry` in `components/Terminal.tsx` used to call `term.reset()` on disconnect + "Start Over", which wipes both the visible screen and the scrollback. Users lost all context from the previous session the moment they clicked retry.
- Replace the `reset()` with the existing `preserveTerminalViewportInScrollback` utility (already used to keep content when the remote sends CSI 2J), plus explicit escape sequences for the modes that actually matter when crossing session boundaries: mouse tracking (1000/1002/1003/1006), alt-screen (1049), SGR colors, and cursor visibility. Also home the cursor.
- Net effect: after a reconnect, the user sees a fresh prompt from the new session but can scroll up to read everything from before the disconnect.

Fixes #695

## Test plan
- [x] SSH into a host, run a few commands (e.g. `ls`, `ps`, `top` then quit).
- [x] Force the connection to drop (e.g. close the network, restart sshd, or kill the tunnel).
- [x] Click "Start Over" / retry.
- [x] After reconnect succeeds, scroll the terminal up and confirm the prior session's output is still visible in scrollback.
- [x] Verify no stale mouse sequences leak as keyboard input (click/drag behaviour stays sane after reconnect).
- [x] Repeat the above while the previous session was in the alternate screen (e.g. disconnect while inside `vim` or `less`). Confirm the new session starts on the main screen, not stuck in alt-screen.
- [x] Repeat for local shell (protocol=local), telnet, serial, and mosh retry paths — buffer should be preserved for all.

### DIFF
--- a/components/Terminal.tsx
+++ b/components/Terminal.tsx
@@ -49,6 +49,7 @@ import { ZmodemProgressIndicator } from "./terminal/ZmodemProgressIndicator";
 import { useZmodemTransfer } from "./terminal/hooks/useZmodemTransfer";
 import { createTerminalSessionStarters, type PendingAuth } from "./terminal/runtime/createTerminalSessionStarters";
 import { createXTermRuntime, primaryFontFamily, type XTermRuntime } from "./terminal/runtime/createXTermRuntime";
+import { preserveTerminalViewportInScrollback } from "./terminal/clearTerminalViewport";
 import { XTERM_PERFORMANCE_CONFIG } from "../infrastructure/config/xtermPerformance";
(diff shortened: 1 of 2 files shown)


binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

PR #693, opened 2026-04-11
fix: Claude ACP agent now uses custom API key and base URL

binaricat:
## Summary

- **Root cause**: The renderer only resolved `openai` providers (for Codex) when selecting a provider ID for ACP agents. Claude agent was never matched, so no `providerId` was passed to the main process. Even if it were, the main process only injected `CODEX_API_KEY` — never `ANTHROPIC_API_KEY` or `ANTHROPIC_BASE_URL`.
- **Renderer fix** (`useAIChatStreaming.ts`): Now uses `matchesManagedAgentConfig` to resolve the correct provider per agent type — `anthropic`/`custom` provider for Claude, `openai` for Codex
- **Main process fix** (`aiBridge.cjs`): Injects `ANTHROPIC_API_KEY` and `ANTHROPIC_BASE_URL` into `claude-agent-acp` env when a matching provider is configured, across all three ACP provider creation paths (list-models, stream, fallback)

This enables users who configure an Anthropic provider with a custom base URL (e.g. CC Switch proxy, LiteLLM, OpenRouter) to use Claude Code without being redirected to the official OAuth flow.

Closes #677

## Test plan

- [x] Configure an Anthropic provider in Settings → AI with a custom API key
- [x] Select the Claude Code agent and send a message — should work without browser auth redirect
- [x] Configure a custom base URL on the provider — verify requests go to the custom endpoint
- [x] Verify Codex still works with its OpenAI provider (no regression)
- [x] Verify Copilot still works (no regression)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #702, opened 2026-04-12
[codex] finish Codex provider follow-up for #677

binaricat:
## Summary

This is a follow-up for #677.

The issue was closed after PR #693, but that fix mainly repaired the Claude path. Codex still had two gaps that made the original report feel only partially fixed:

- Codex ACP only resolved enabled `openai` providers, so users relying on the app's OpenAI-compatible `custom` provider could still fall back to the ChatGPT login path.
- The ACP provider reuse fingerprint did not include Codex provider/base URL state, so changing the provider or endpoint inside an existing chat session could keep reusing a stale Codex ACP instance.

## What changed

- Added a shared managed-agent provider resolver and used it for Codex/Claude provider selection.
- Codex now prefers an enabled `openai` provider, but falls back to an enabled OpenAI-compatible `custom` provider with a base URL.
- Updated the Codex settings hint text so it reflects OpenAI-compatible providers instead of only the OpenAI preset.
- Included provider identity and base URL in the ACP auth fingerprint so Codex/Claude recreate the ACP provider when credentials or endpoint settings change.

## Why

The original issue asked for Codex to work with custom model endpoints such as CC Switch. The previous fix chain improved part of that story, but Codex still had leftover edge cases that could keep redirecting users to the login flow or keep using stale endpoint settings.

## Validation

- `npm run lint`
- `npm run build`
- `node -c electron/bridges/aiBridge.cjs`
- local provider-selection check covering:
  - Codex prefers `openai`
  - Codex falls back to `custom`
  - Claude still prefers `anthropic`


binaricat:
@codex review

PR #717, opened 2026-04-14
[codex] Fix mac fullscreen close-to-tray behavior

binaricat:
## Summary
- fix the macOS close-to-tray path so closing a fullscreen window leaves fullscreen before hiding the app window
- keep the existing close behavior after leaving fullscreen instead of turning the close button into a fullscreen toggle
- add regression coverage for normal close, tray close, and fullscreen tray close flows

## Root Cause
When close-to-tray was enabled, clicking the window close button always hid the main window immediately. On macOS, doing that while the window was in native fullscreen could leave behind an empty fullscreen Space, which showed up as a black or blank screen.

## Validation
- `node --test electron/bridges/globalShortcutBridge.test.cjs`
- `npm run lint`
- `npm run build`

Fixes #716

PR #718, opened 2026-04-14
fix: stop cancelling mac fullscreen tray-hide on internal show event

binaricat:
## Summary

Follow-up to #717. On macOS, clicking the red close button on a fullscreen window with close-to-tray enabled was exiting fullscreen but **leaving the window visible** instead of hiding it to the tray.

## Root cause

\`hideWindowRespectingMacFullscreen\` registered three \"cancel pending hide\" listeners in PR #717:

\`\`\`js
win.once(\"leave-full-screen\", ...);   // correct — triggers hide
win.once(\"closed\", ...);              // correct — window gone, drop pending
win.once(\"show\", ...);                // ← problem
\`\`\`

macOS fires a \`show\` event on the BrowserWindow **internally** while the native fullscreen exit animation lands the window back in its home Space. That is not user intent, but the listener treated it as such and cleared the pending hide (including the \`leave-full-screen\` listener), so the hide never happened.

## Fix

- Remove the \`show\` listener entirely. All the legitimate \"bring the window back during the exit animation\" entry points (\`openMainWindow\`, \`toggleWindowVisibility\`, \`setCloseToTray(false)\`, the tray \"Open Main Window\" menu) already call \`clearPendingFullscreenHide\` explicitly, so the listener was only ever catching the internal transition emit.
- Export \`clearPendingFullscreenHide\` from the bridge and call it from \`app.on(\"activate\")\` in \`main.cjs\` — so a dock-click during the exit animation still cancels the pending hide cleanly (user intent > pending hide).

## Tests

- Renamed \`show event cancels a pending fullscreen hide\` → \`show event does **not** cancel a pending fullscreen hide\`, asserting the pending state survives a spurious \`show\` and still hides on \`leave-full-screen\`.
- Added \`app activate clears a pending fullscreen hide\` covering the new \`clearPendingFullscreenHide\` bridge export.
- Dropped \`listenerCount(\"show\")\` assertions from the other tests (the listener no longer exists).

All 12 tests in \`globalShortcutBridge.test.cjs\` pass. Lint clean.

## Validation

- [x] \`node --test electron/bridges/globalShortcutBridge.test.cjs\`
- [x] \`npm run lint -- --quiet\`
- [ ] Manual: on macOS, fullscreen a window with close-to-tray enabled → click red close → window exits fullscreen AND hides to the tray (no pop-back)
- [ ] Manual: same setup, then during the fullscreen exit animation click the dock icon → window comes back visible (pending hide cancelled)

Related: #716, #717

### DIFF
--- a/electron/bridges/globalShortcutBridge.cjs
+++ b/electron/bridges/globalShortcutBridge.cjs
@@ -51,7 +51,6 @@ function clearPendingFullscreenHide(win) {
   try {
     win.removeListener?.("leave-full-screen", pending.onLeaveFullScreen);
     win.removeListener?.("closed", pending.onClosed);
-    win.removeListener?.("show", pending.onShow);
   } catch {
     // ignore
   }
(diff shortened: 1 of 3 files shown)


binaricat:
@codex review

PR #723, opened 2026-04-14
fix: address Codex follow-up review on PR #720

binaricat:
## Summary
Address the two issues Codex's automated review flagged on #720 after merge.

- **P1** (useAutoSync): open the auto-sync gate when startup retries are exhausted, instead of leaving the session silently non-syncing until restart.
- **P2** (CloudSyncSettings): disable ALL restore buttons while any restore is running, not just the clicked row.

## Why
Codex reviewed commit `8b85759816` post-merge and surfaced:

### P1 — retry exhaustion wedges auto-sync
The retry effect returned at \`attempt >= 4\` without setting \`remoteCheckDoneRef\`. A session with a persistent inspect failure (long outage, provider rate-limit) had auto-sync silently disabled for the rest of the session — edits accumulated locally, nothing pushed, user had no signal beyond the one toast that fired at minute zero.

The original reason for gate-closing on inspect failure was to prevent **pushing partial/empty state over remote**. That's still important, but in the interim it's now covered by independent guards:
- \`hasMeaningfulSyncData\` in \`syncNow\` refuses to push an empty vault on auto triggers.
- The apply-in-progress sentinel refuses to push a half-applied vault.
- \`checkProviderConflict\` runs before every upload and throws on inspect failure, giving per-push safety even without a successful startup inspect.

After retry exhaustion, opening the gate matches the semantic that manual sync ALWAYS had (push is allowed without a completed startup reconcile), rather than making auto-sync stricter than the manual path.

### P2 — concurrent restore clicks can interleave
\`applyProtectedSyncPayload\` is NOT serialized same-window:
- \`withRestoreBarrier\` writes a cross-window deadline; two same-window calls both write it, both run their applies, both clear. Cross-window safe, same-window not.
- Each run writes an apply-in-progress sentinel at start and clears it at end. Two overlapping runs: clear from run B masks the fact that run A is still partial → next-session crash detection thinks apply finished cleanly.

Disable every row when \`restoringBackupId !== null\` (not just the clicked row) so the user physically cannot start a second concurrent restore.

## What changed
- \`application/state/useAutoSync.ts\` — P1 fix + comment explaining the independent guards that make gate-opening safe here.
- \`components/CloudSyncSettings.tsx\` — P2 fix: \`disabled={restoringBackupId !== null}\` on the per-row restore button.

## Test plan
- [x] \`npm run lint\` clean
- [x] \`npm run build\` clean
- [x] \`node --test ...\` 50/50 pass
- [ ] Manual: force an inspect-failure state for > 5 min, confirm an edit still auto-syncs
- [ ] Manual: click two restore buttons rapidly, confirm the second is ignored

### DIFF
--- a/application/state/useAutoSync.ts
+++ b/application/state/useAutoSync.ts
@@ -284,6 +284,18 @@ export const useAutoSync = (config: AutoSyncConfig) => {
       }
 
       lastSyncedDataRef.current = dataHash;
+
+      // Successful sync implies a successful per-provider
+      // `checkProviderConflict` (which inspects remote) — equivalent
+      // to a successful startup reconciliation from the auto-sync
+      // gate's point of view. Opening the gate here is the escape
+      // hatch when a network outage exhausted the startup retry
+      // timer: a user-triggered manual sync (or any first successful
+      // auto sync that somehow ran anyway) resumes auto-sync for the
+      // rest of the session. Without this, a degraded-startup session
+      // would require the user to manually sync after every edit.
+      hasCheckedRemoteRef.current = true;
+      remoteCheckDoneRef.current = true;
     } catch (error) {
       if (trigger === 'manual') {
         throw error;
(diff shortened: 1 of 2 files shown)


binaricat:
@codex review

PR #724, opened 2026-04-15
fix: stabilize scoped AI draft/session transitions

tces1:
## Summary

This PR hardens scoped AI chat state handling across terminal and workspace panels.

It separates per-scope draft state from persisted session history, tightens panel view/session restore behavior, improves cleanup for closed scopes, and fixes a draft attachment race where normal draft edits could invalidate in-flight uploads.

## What changed

- moved AI draft composer state and panel view state into explicit per-scope transient state
- added dedicated helpers for:
  - draft lifecycle and panel view transitions
  - inactive scope cleanup
  - panel view normalization and displayed-session resolution
  - draft send gating
  - terminal/workspace session scope matching
- changed terminal AI behavior to start from a fresh draft by default instead of auto-resuming history into a newly opened terminal
- preserved workspace history resume behavior, including persisted active-session restoration when available
- cleaned up orphaned scoped state more consistently when terminals/workspaces close
- extracted file-to-upload conversion into shared helpers and fixed upload invalidation logic by tracking upload lifecycle separately from normal draft mutations
- improved AI session history drawer layout behavior
- updated managed-agent behavior so Claude/Codex auth follows CLI-owned login/config instead of renderer-side provider assumptions
- removed outdated Codex API-key hint copy that no longer reflects the current integration model

## Why

The previous flow mixed persisted session state with transient draft/panel intent, which caused several edge-case regressions:

- stale draft/view state could survive longer than intended
- terminal history could resume into the wrong context
- closing scoped targets could leave behind orphaned transient state
- file uploads could be dropped if the user edited the draft while conversion was still in flight

This PR makes those state boundaries explicit and narrows the restore rules so each scope behaves predictably.

## Testing

- `node --test application/state/aiDraftState.test.ts application/state/aiScopeCleanup.test.ts components/ai/aiPanelViewState.test.ts components/ai/draftSendGate.test.ts components/ai/sessionHistoryLayout.test.ts components/ai/sessionScopeMatch.test.ts`
- `npm run lint`
- `npm run build`

## Notes

There are also local uncommitted changes in:
- `application/state/aiDraftState.ts`
- `application/state/useAIState.ts`
- `application/state/aiDraftState.test.ts`

Those changes further harden the draft upload race by separating upload generation from general draft mutation tracking. If they will be included before merge, they should be considered part of this PR's final scope.

### DIFF
--- a/application/state/aiDraftState.test.ts
+++ b/application/state/aiDraftState.test.ts
@@ -3,9 +3,13 @@ import assert from "node:assert/strict";
 
 import {
   activateDraftView,
+  bumpDraftMutationVersionState,
+  bumpDraftUploadGenerationState,
   clearScopeDraftState,
   createEmptyDraft,
   ensureDraftForScopeState,
+  getDraftMutationVersionState,
+  getDraftUploadGenerationState,
   pruneTerminalScopeState,
   pruneTerminalTransientState,
   resolvePanelView,
@@ -166,6 +170,28 @@ test("ensureDraftForScopeState returns the original ref when the scope already e
   );
 
   assert.equal(next, draftsByScope);
+});
+
+test("draft mutation version increments on every mutation for the same scope", () => {
+  const scopeKey = "terminal:1";
+  const initialVersion = getDraftMutationVersionState({}, scopeKey);
+  const nextVersions = bumpDraftMutationVersionState({}, scopeKey);
+  const finalVersions = bumpDraftMutationVersionState(nextVersions, scopeKey);
+
+  assert.equal(initialVersion, 0);
+  assert.equal(getDraftMutationVersionState(nextVersions, scopeKey), 1);
+  assert.equal(getDraftMutationVersionState(finalVersions, scopeKey), 2);
+});
+
+test("draft upload generation only increments when the draft lifecycle rolls over", () => {
+  const scopeKey = "terminal:1";
+  const initialGeneration = getDraftUploadGenerationState({}, scopeKey);
+  const nextGenerations = bumpDraftUploadGenerationState({}, scopeKey);
+  const finalGenerations = bumpDraftUploadGenerationState(nextGenerations, scopeKey);
+
+  assert.equal(initialGeneration, 0);
+  assert.equal(getDraftUploadGenerationState(nextGenerations, scopeKey), 1);
+  assert.equal(getDraftUploadGenerationState(finalGenerations, scopeKey), 2);
 });
 
 test("pruneTerminalScopeState removes closed terminal drafts and views only", () => {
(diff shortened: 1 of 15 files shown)


binaricat:
@codex review

binaricat:
Pushed a follow-up commit (345244b2) that addresses the latest Codex P2 on `aiPanelViewState.ts:38`.

**What it does**

- Added a dedicated effect in `AIChatSidePanel` that clears `activeSessionIdMap[scopeKey]` whenever the resolved panel view is `draft` but a persisted session id is still present. This keeps the map an accurate record of "which session each scope is currently displaying" instead of a lagging snapshot, so `activeTerminalTargetIds` no longer falsely claims ownership of a session that isn't being shown anymore.
- Extended `sessionScopeMatch.test.ts` with the rank=2 exact-match case and the scope-type mismatch short-circuit, both of which were missing from the original suite.

`node --test components/ai/sessionScopeMatch.test.ts` passes; no TS errors on the touched files.

Hope you don't mind me pushing directly — `maintainerCanModify` is on, so I figured this was quicker than round-tripping through a comment. Happy to revert if you prefer to do it yourself.

@codex review

binaricat:
@codex review

binaricat:
Pushed commit `8d8b39c5` that addresses the latest Codex P2 on `AIChatSidePanel.tsx:382`.

**What it does**

Switched cross-terminal session-ownership tracking from `session.scope.targetId` to `session.id`, since the stored `scope.targetId` is a creation-time fact and no longer reflects "which terminal is currently displaying this session" once the user resumes a session from history into a different terminal.

Concretely:

- `getSessionScopeMatchRank`'s 5th parameter is renamed `activeTerminalTargetIds` → `activeTerminalSessionIds`; the ownership check becomes `activeTerminalSessionIds?.has(session.id)`. This prevents the same session from being resumed simultaneously in multiple terminals.
- `AIChatSidePanel` derives that set from the *values* of `activeSessionIdMap` (the session ids live scopes are showing), skipping the current scope key.
- `pruneInactiveScopedSessions` gains an `activeSessionIds` parameter. A session whose id is in the in-use set is no longer reported as orphaned and its `externalSessionId` is no longer cleared, even when its stored `scope.targetId` points at a closed terminal. Protects ACP continuity for history sessions resumed into a different terminal.
- `cleanupOrphanedAISessions` computes the in-use set from the pre-cleanup `activeSessionIdMap` (filtered to live scopes) and passes it through. Map is now read once instead of twice.

**Tests**

Extended `sessionScopeMatch.test.ts` and `aiScopeCleanup.test.ts`; the new cases cover id-based ownership, exact-match rank=2, scope-type mismatch short-circuit, and the "resumed-elsewhere session must not be cleaned" invariant. Full node --test run on both files passes (9 tests). `tsc --noEmit` clean on the touched files.

@codex review

PR #726, opened 2026-04-15
feat(ai-chat): fit-to-content popovers + keyboard nav for @/slash menus

binaricat:
## Summary

- The `@` host picker and `/skill` picker in the AI chat input now shrink to fit their content (min 220px, max = input width) instead of always matching the full input width. Short lists no longer leave a wide empty box.
- Added keyboard navigation to both popovers:
  - `↑` / `↓` cycle through items (wrap at ends)
  - `Enter` commits the highlighted item; does not fall through to message submit while a menu is open
  - `Esc` closes the menu
  - Mouse hover syncs the active index so keyboard and pointer never disagree
- Exposed `aria-selected` / `aria-activedescendant` on the listbox for assistive tech.

## Test plan

- [ ] Type `@` in the chat input with ≥1 connected host — popover hugs the host rows.
- [ ] Use `↑`/`↓` to move the highlight, press `Enter` to insert `@hostname ` into the input (no message submit).
- [ ] Press `Esc` to close the menu; typing resumes normally.
- [ ] Hover a different row with the mouse; pressing `Enter` now selects that row.
- [ ] Repeat for `/skill` popover (type a leading `/`): same width/keyboard behavior.
- [ ] Before/after screenshots on a single-host list to confirm the width is tighter.

### DIFF
--- a/components/ai/ChatInput.tsx
+++ b/components/ai/ChatInput.tsx
@@ -7,7 +7,7 @@
  */
 
 import { AtSign, Check, ChevronDown, ChevronRight, Cpu, Expand, Eye, FileText, ImageIcon, Package, Plus, ShieldCheck, X, Zap } from 'lucide-react';
-import React, { useCallback, useRef, useState } from 'react';
+import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react';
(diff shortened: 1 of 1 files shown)


binaricat:
@codex review

PR #733, opened 2026-04-16
[codex] Accept empty SSH identification banners for compatibility

### LINKED ISSUE #730: 某些ssh版本的服务器无法正常连接
版本：1.0.92
问题：ssh server 版本为“SSH-2.0-”的服务器。提示[Failed to start SSH: Error invoking remote method 'netcatty:start': Error: Invalid identification string],这个应该是使用的ssh客户端组件有校验ssh server版本格式导致的。

binaricat:
## Summary
- accept remote SSH banners like `SSH-2.0-` instead of failing during header parsing
- keep the validation narrow so malformed banners are still rejected
- add a regression test covering empty, normal, and malformed identification strings

## Root cause
Netcatty uses `ssh2` for SSH transport. Its header parser required the `softwareversion` token in the remote identification string to be non-empty, so servers that send `SSH-2.0-` failed immediately with `Invalid identification string` before authentication even started.

## Notes
This is a targeted compatibility exception for non-compliant servers.

- The SSH spec still defines `softwareversion` as non-empty: [RFC 4253 §4.2](https://www.rfc-editor.org/rfc/inline-errata/rfc4253.html#section-4.2)
- OpenSSH also rejects empty software versions in banner parsing: [openssh-portable `kex.c`](https://github.com/openssh/openssh-portable/blob/b433d015a6547598e568808d9cae16325bc8928f/kex.c)

So this patch only relaxes the empty-token case and keeps other malformed banners rejected.

Closes #730

## Validation
- `node --test /Users/chenqi/projects/personal/netcatty/electron/bridges/sshIdentificationCompatibility.test.cjs`
- `npm run build`

### DIFF
--- a/electron/bridges/sshIdentificationCompatibility.test.cjs
+++ b/electron/bridges/sshIdentificationCompatibility.test.cjs
@@ -1,0 +1,46 @@
+const test = require("node:test");
+const assert = require("node:assert/strict");
+
+const Protocol = require("ssh2/lib/protocol/Protocol");
+
+function parseIdentification(line) {
+  let header;
+  const protocol = new Protocol({
+    onWrite() {},
+    onError(err) {
+      throw err;
+    },
+    onHeader(nextHeader) {
+      header = nextHeader;
+    },
+  });
+
+  const data = Buffer.from(`${line}\r\n`, "latin1");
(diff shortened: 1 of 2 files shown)


binaricat:
@codex review

PR #743, opened 2026-04-16
Tab bar: duplicate-adjacent insertion + wheel-to-horizontal scroll

binaricat:
## Summary
Addresses #737.

- **Duplicate tab insertion**: `copySession` now inserts the new session id immediately after the source in `tabOrder`, so duplicated tabs appear next to the original instead of being appended to the far right (which was hard to find with many tabs open).
- **Wheel-to-horizontal scroll**: the top tab strip now translates vertical mouse-wheel deltas into horizontal scroll, so users with many tabs can reach either end of the strip without dragging. Trackpad gestures that already carry horizontal delta are passed through untouched to preserve native two-finger swiping.

## Test plan
- [x] With several tabs open, right-click a middle tab → *Duplicate*; the new tab should appear directly to the right of the source (not at the end).
- [x] Duplicating a tab still activates the new tab.
- [x] Open enough tabs to cause the strip to overflow; scroll the mouse wheel over the tab bar — the strip should scroll horizontally in the wheel direction.
- [x] Two-finger horizontal swipe on a trackpad still scrolls horizontally as before.
- [x] Vertical page/content scroll in other areas is unaffected.
- [x] Drag-reordering tabs still works.

### DIFF
--- a/application/state/useSessionState.ts
+++ b/application/state/useSessionState.ts
@@ -668,37 +668,55 @@ export const useSessionState = () => {
   const copySession = useCallback((sessionId: string, options?: {
     localShellType?: TerminalSession['shellType'];
   }) => {
-    setSessions(prevSessions => {
-      const session = prevSessions.find(s => s.id === sessionId);
-      if (!session) return prevSessions;
-      const nextShellType = session.protocol === 'local'
-        ? options?.localShellType
-        : session.shellType;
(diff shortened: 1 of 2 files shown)


binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

PR #769, opened 2026-04-18
Preload compact history on first turn after app restart (#753 hedge)

binaricat:
## Summary

Actual fix for #753 (and the symptom shown in the recent screenshot on GitHub Copilot CLI): after closing and reopening Netcatty, the AI chat UI shows the prior conversation but the agent responds \"this is the beginning of our conversation\". Earlier context is lost.

PR #754 only replayed history when the ACP layer threw an explicit \"session not found\" error. Some ACP agents (Copilot CLI, some Codex builds) **silently** spawn a new session when handed a stale id — no error → no fallback → first turn sends just the latest prompt → agent has zero context.

## Fix

When we're creating a new provider process AND passing it an \`existingSessionId\` AND the renderer gave us compact history, preload \`historyReplayFallback = true\` **as a hedge**:

- If the agent really did reload the session: replay is ~3KB of redundant context — small waste
- If the agent silently started fresh: replay restores durable constraints + the last few raw turns, so the first response is coherent

After the first successful streamed turn, the existing post-stream hook clears the flag. Steady state is back to sending only the latest prompt. **One replay per app-restart-and-prompt.**

## Why we can't just detect silent-resume

There's no ACP-level signal that distinguishes \"session resumed with full state\" from \"session created fresh under the old id\". The only indicator is the agent's actual first response, which happens *after* we've already built the streamText call. Hedging up front is the only correct strategy.

## Files

| File | Change |
|---|---|
| \`electron/bridges/aiBridge.cjs\` | Extend \`preserveHistoryReplayFallback\` with the \"new provider + resumeSessionId + historyMessages\" clause |
| \`electron/bridges/aiBridge.test.cjs\` | New test: \"replays compact history on the first turn after app restart even when session/load 'succeeds'\" + parameterize mock \`createACPProvider\` |

## Test plan

- [x] Reproduce #753: connect an external agent, have a few turns, close Netcatty, reopen, select the prior AI session, send a message that references earlier context (\"what did we discuss?\") — agent should now see prior content
- [x] Fresh chat (no \`existingSessionId\`) — single provider creation, no replay (no regression)
- [x] Steady-state conversation (multiple turns after restart) — second+ turns send only latest prompt (flag clears after first successful turn)

Hedges #753 (cannot fully \"close\" without broader ACP protocol changes, but this is the material fix for the observed symptom).

### DIFF
--- a/application/state/aiDraftState.test.ts
+++ b/application/state/aiDraftState.test.ts
@@ -13,6 +13,7 @@ import {
   pruneTerminalScopeState,
   pruneTerminalTransientState,
   resolvePanelView,
+  selectDraftForAgentSwitch,
   setDraftView,
   setSessionView,
   updateDraftForScope,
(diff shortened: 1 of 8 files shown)


binaricat:
@codex review

binaricat:
@codex review

PR #772, opened 2026-04-19
[codex] avoid main-process runtime crashes

binaricat:
## What changed
- switched the main-process crash policy to distinguish startup failures from runtime failures
- keep startup errors fatal, but suppress runtime uncaught exceptions and unhandled rejections after the main window is up
- moved the process-error rules into a dedicated helper and added regression tests for startup-vs-runtime behavior and Chromium-style network errors

## Why
Issue #771 showed that a transient network change could surface as an unhandled rejection in the main process and bring down the whole app. The deeper problem was that runtime failures were still treated like startup failures, so a single background task error could terminate Netcatty.

## Impact
- Netcatty now stays open when a runtime task crashes or a background promise rejects after startup
- startup remains strict, so broken boot paths still fail loudly instead of hiding a bad launch
- the behavior is covered by regression tests so we do not drift back to the old crash policy

## Validation
- `node --test electron/bridges/mainProcessErrorGuards.test.cjs`
- `npm test`
- `node --check electron/main.cjs`

### DIFF
--- a/electron/bridges/mainProcessErrorGuards.test.cjs
+++ b/electron/bridges/mainProcessErrorGuards.test.cjs
@@ -1,7 +1,10 @@
 const test = require("node:test");
 const assert = require("node:assert/strict");
+const { EventEmitter } = require("node:events");
 const {
   classifyProcessError,
+  createProcessErrorController,
+  installProcessErrorHandlers,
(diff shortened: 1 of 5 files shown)


binaricat:
@codex review

PR #784, opened 2026-04-21
Block empty/shrunk pushes when sync base is null (#779)

### LINKED ISSUE #779: Bug: 【1.0.95版本】mac上的钥匙串，win11同步之后，win11和MAC的钥匙串全部消失。
## Describe the problem
1、mac上的钥匙串，win11同步之后，win11和MAC端的钥匙串全部消失。
2、win11我手动配置的钥匙串也没了。
3、我是OneDrive同步的
## Steps to reproduce
1.mac点击同步成功之后
2.windows11点点击同步，
2.1 win11上和MAC的钥匙串全没了。

Image

## Environment
- App: netcatty 1.0.95
- Platform: win32
- UA: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) netcatty/1.0.95 Chrome/144.0.7559.96 Electron/40.1.0 Safari/537.36

binaricat:
## Summary

Root cause of #779 + three related sync-robustness gaps surfaced during the investigation. All four land in this PR since they share test surface.

### 1. Shrink guard ignored null base (primary #779 root cause)

`detectSuspiciousShrink` short-circuited to `{suspicious:false}` whenever `base === null`. That happens on:

- First sync on a new install (no base yet)
- Post-unlock window where `unlockedKey` was re-derived and `loadSyncBase` returns null
- Cases where the encrypted base blob fails to decrypt

Combined with OneDrive treating a not-yet-delivered remote file as 404 (see #2) and timing races where a fresh device's local keychain hasn't fully hydrated, a device in this state would quietly push an empty payload that overwrites the populated cloud copy. The other device then pulled the empty cloud, merge concluded "both sides deleted", and the data was gone on both ends — exactly what the reporter experienced (Mac → OneDrive → Win11 keychain wipe).

**Fix:** `domain/syncGuards.ts` accepts an optional `remote` fallback. When `base` is null, the guard compares against the remote payload instead. Three call sites in `CloudSyncManager.ts` now pass the already-decrypted remote (merge branch) or decrypt `checkResult.remoteFile` on demand (direct-upload and `syncAllProviders`). `ShrinkFinding` gains an optional `viaRemote` marker.

Legitimate scenarios remain unaffected:
- No base **and** no remote → still `not suspicious` (genuinely empty cloud).
- Outgoing larger than remote (first-sync growth) → `lost` is negative, guard skips.
- Base present → behaviour unchanged.

### 2. OneDrive 404 eventual-consistency retry

`findSyncFile` / `downloadSyncFile` in `OneDriveAdapter.ts` treated every `null`/404 result as authoritative "cloud is empty". That was a critical step in how #779 escalated: the second device sees 404 before OneDrive has propagated the upload, decides the cloud is empty, and proceeds to overwrite.

**Fix:** both functions now retry on null results with 1.5s / 3s backoff (bounded to 2 extra attempts). Non-404 error paths are unchanged.

### 3. Restore-barrier self-heal

`isRestoreInProgress` in `useAutoSync.ts` trusts the stored epoch-ms deadline without sanity-checking it. The heartbeat-based `withRestoreBarrier` already handles the normal crash case (deadline naturally expires within ~60s), but a stale past-deadline would linger in localStorage and a pathological future-deadline could wedge auto-sync indefinitely.

**Fix:** when `isRestoreInProgress()` sees a deadline in the past, clear the key. When it sees a deadline more than 10 minutes in the future, treat as corrupt, log, and clear.

### 4. Multi-provider divergence now emits an event

The existing divergent-provider-bases detection in `syncAllProviders` only logged a `console.warn` — the UI had no way to notice that uploading a merged payload would overwrite a differently-configured provider's data.

**Fix:** added a `PROVIDERS_DIVERGED` member to `SyncEvent`, emitted next to the existing `console.warn`. UI consumers can subscribe via the existing `manager.subscribe` path; no other event shapes changed.

## Test plan

- [x] `domain/syncGuards.test.ts` — 4 new cases (null base + various remote states), all 17 pass.
- [x] `tsc --noEmit` clean on all touched files.
- [ ] Reproduce #779: fresh Win11 with OneDrive, Mac with populated keychain → Win11's first sync is blocked by the shrink banner instead of pushing empty.
- [ ] Legitimate first sync: fresh install with real local data, empty cloud → upload succeeds (no block).
- [ ] OneDrive race: rapid successive syncs from two devices → no data loss; second device sees the file via retry instead of 404-as-empty.
- [ ] Kill app mid-restore → auto-sync unblocks within the normal hold window; no lingering localStorage key next time `isRestoreInProgress` runs.
- [ ] Connect two providers pointed at different accounts → a `PROVIDERS_DIVERGED` event fires on the next conflict-triggering sync.

Closes #779

### DIFF
--- a/application/state/useAutoSync.ts
+++ b/application/state/useAutoSync.ts
@@ -51,10 +51,35 @@ const AUTO_SYNC_PROVIDER_ORDER: CloudProvider[] = ['github', 'google', 'onedrive
 
 // Cross-window restore barrier: stored as an epoch-ms deadline. Any value
 // in the future means a restore is applying in some window and auto-sync
-// must not push concurrently.
+// must not push concurrently. The writer (`withRestoreBarrier`) heartbeats
+// the deadline to keep it alive; a crashed window naturally expires within
+// ~RESTORE_BARRIER_HOLD_MS. We still defend against two degenerate cases:
+// (1) a stale deadline sitting in the past — harmless but pollutes debug
+//     state, so we opportunistically clear it; (2) a deadline absurdly far
+//     in the future (clock skew between windows, pathological holdMs, or a
+//     tampered value) — would otherwise lock auto-sync indefinitely, so we
+//     clear it and treat the barrier as inactive.
+const RESTORE_BARRIER_SANITY_MAX_MS = 10 * 60 * 1000; // 10 minutes
 const isRestoreInProgress = (): boolean => {
   const raw = localStorageAdapter.readNumber(STORAGE_KEY_VAULT_RESTORE_IN_PROGRESS_UNTIL);
-  return typeof raw === 'number' && raw > Date.now();
+  if (typeof raw !== 'number' || raw <= 0) return false;
+  const now = Date.now();
+  if (raw <= now) {
+    // Deadline is in the past — either a clean finish that failed to
+    // overwrite the key, or a crashed heartbeat. Clear so subsequent
+    // reads are cheap and the key doesn't linger forever.
+    localStorageAdapter.writeNumber(STORAGE_KEY_VAULT_RESTORE_IN_PROGRESS_UNTIL, 0);
+    return false;
+  }
+  if (raw - now > RESTORE_BARRIER_SANITY_MAX_MS) {
+    console.warn(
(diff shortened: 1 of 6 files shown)


binaricat:
@codex review

PR #785, opened 2026-04-21
Vault global search spans all groups/packages (#777)

### LINKED ISSUE #777: [Feature Request] ：希望vaults页面全局检索支持检索分组内的内容
当前现象：在vaults下不管是主机还是代码片段页面全局搜索功能无法检索到被添加到分组（并开启根目录只显示未分组主机时）或者代码包中的内容。
期望行为：在检索栏输入关键字后索引的范围不知包含但前页面，也包含当前页面的分组或代码包中。

binaricat:
## Summary

Search was scoped to the current group (hosts page) or the current package (snippets page). If a host lived in a group the user hadn't drilled into — or if "root only shows ungrouped hosts" was turned on — a matching name would silently not appear in the search results. Report: #777.

### Hosts (`VaultView.tsx`)
- `displayedHosts` now skips the `selectedGroupPath` / `showOnlyUngroupedHostsInRoot` branches when the search box is non-empty.
- Each matching card shows a small outline badge with the host's group so cross-group results are recognizable.
- `treeViewHosts` already did this — the change makes the flat grid/list views consistent with the tree.

### Snippets (`SnippetsManager.tsx`)
- `displayedSnippets` skips the `selectedPackage` filter when search is active.
- Sub-package tile grid is hidden while searching (would be redundant alongside a flat cross-package list).
- Each snippet card shows the package path as a small bordered tag.

Empty state, sort, and tag filters are unchanged.

Closes #777

## Test plan
- [ ] Enable "root only shows ungrouped hosts", type a name that's in a group → host appears with its group badge.
- [ ] Inside a sub-group, search for a host that lives in a sibling group → appears.
- [ ] Clear search → list returns to the current group's scope as before.
- [ ] Snippets: navigate into a package, search for a snippet in another package → appears with package badge; sub-package tiles hide while search active.
- [ ] Clear snippet search → package tiles and current-package scoping return.

### DIFF
--- a/application/i18n/locales/en.ts
+++ b/application/i18n/locales/en.ts
@@ -1682,6 +1682,8 @@ const en: Messages = {
   'snippets.breadcrumb.separator': '›',
   'snippets.empty.title': 'Create snippet',
   'snippets.empty.desc': 'Save your most used commands as snippets to reuse them in one click.',
+  'snippets.search.noResults.title': 'No matches',
+  'snippets.search.noResults.desc': 'No snippets or packages match "{query}". Try a different search term or clear the search to browse.',
   'snippets.section.packages': 'Packages',
   'snippets.section.snippets': 'Snippets',
   'snippets.package.count': '{count} snippet(s)',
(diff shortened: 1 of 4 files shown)


binaricat:
Addressed in b970f225: added a dedicated no-results panel (title + description, zh-CN & en) that renders when `search.trim()` has no snippet or package match but snippets still exist. Package tiles stay hidden during search as before; the user now sees feedback + a hint to clear search instead of a blank area. @codex

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

PR #788, opened 2026-04-21
Polish workspace focus-mode sidebar

binaricat:
## Summary

Round of polish on the focus-mode workspace sidebar (the \"Terminals · N\" session list that shows inside a Workspace tab). Four distinct fixes that all landed together in this commit:

### 1. Stop syncing position with the terminal side panel

Previously the outer row used \`flex-row-reverse\` to push the terminal side panel to the right, which also flipped the focus-mode sidebar to the right. The workspace sidebar should always be anchored to the left regardless of the side panel's side.

Replaced \`flex-row-reverse\` with \`order-last\` on the side panel itself, so the focus-mode sidebar and terminal area stay in source order. Focus sidebar → terminal → side panel (left/right as configured).

### 2. Drag-resizable width

New persisted width with storage key \`STORAGE_KEY_WORKSPACE_FOCUS_SIDEBAR_WIDTH\`, default \`224px\` (same as the old \`w-56\`), clamped 160..480. Drag handle on the right edge matches the terminal side panel's resize affordance — \`rAF\`-throttled \`mousemove\`, persisted on \`mouseup\`.

### 3. Paint with the terminal's theme background

The sidebar used to tint itself with \`bg-secondary/50\`, which stood out against the terminal output area as a distinct panel. Switched to inline \`backgroundColor: resolvedPreviewTheme.colors.background\` (and \`color\` for the foreground) so it reads as one continuous surface with the focused terminal. The \`border-r\` stays as a thin separator from the terminal column.

### 4. Restyle session rows

Session rows are now \`RippleButton\`s (same click-ripple feel as the Vault sidebar) instead of plain \`<div>\`s, with the primary-tinted selection treatment swapped out for neutral greys on top of the terminal-theme bg:

- **Selected**: \`bg-foreground/10 text-foreground hover:bg-foreground/15\`, hostname \`font-semibold\`.
- **Unselected**: \`bg-transparent text-foreground/75 hover:bg-foreground/10\`, hostname \`font-medium\`.
- \`hover:text-inherit\` pins the text color against Button's ghost-variant \`hover:text-accent-foreground\`, so the title doesn't flip colour when the cursor passes over a row.
- Dropped the old \`border border-primary/30\` selection outline and the primary tint entirely.

## Test plan

- [ ] Open a Workspace with 2+ terminals, toggle Focus Mode from the split-view switcher.
- [ ] Pin the terminal side panel to the right — focus sidebar stays on the left.
- [ ] Drag the focus sidebar's right edge — width changes smoothly, persists after refresh, clamped at 160 / 480.
- [ ] With a dark terminal theme, the sidebar bg is visually indistinguishable from the terminal output area; selected row shows as a subtle neutral highlight.
- [ ] Hover over an unselected row — only the bg overlay changes; title colour, font size, and font weight stay put.
- [ ] Click switches selection; ripple effect plays, consistent with Vault sidebar entries.
- [ ] Switch to a light terminal theme — sidebar bg follows; text stays readable.

### DIFF
--- a/components/TerminalLayer.tsx
+++ b/components/TerminalLayer.tsx
@@ -29,7 +29,10 @@ import { cn, normalizeLineEndings } from '../lib/utils';
 import { detectLocalOs } from '../lib/localShell';
 import { useStoredString } from '../application/state/useStoredString';
 import { useStoredNumber } from '../application/state/useStoredNumber';
-import { STORAGE_KEY_SIDE_PANEL_WIDTH } from '../infrastructure/config/storageKeys';
+import {
+  STORAGE_KEY_SIDE_PANEL_WIDTH,
+  STORAGE_KEY_WORKSPACE_FOCUS_SIDEBAR_WIDTH,
+} from '../infrastructure/config/storageKeys';
 import { buildCacheKey } from '../application/state/sftp/sharedRemoteHostCache';
 import type { DropEntry } from '../lib/sftpFileUtils';
 import { GroupConfig, Host, Identity, KnownHost, SSHKey, Snippet, TerminalSession, TerminalTheme, Workspace, WorkspaceNode } from '../types';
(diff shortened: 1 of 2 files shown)


binaricat:
@codex review

PR #801, opened 2026-04-22
Render snippets sidebar as an expandable tree (#800)

### LINKED ISSUE #800: 代码片段侧栏可以支持树形结构吗，多个代码包里的片段，需要来回点进点出
代码片段侧栏可以支持树形结构吗，多个代码包里的片段，需要来回点进点出
或者可以鼠标侧键退回上一层，现在退回上一层只能点左上角

binaricat:
## Summary
Closes #800 (approximately — also covers the "鼠标侧键退回上一层" alternative by removing the need to navigate in/out entirely).

The terminal's scripts side panel used breadcrumb navigation, so crossing packages meant clicking out and back in. Replace that with a single tree view where every package row has a chevron to expand/collapse (same affordance as the SFTP tree), so snippets across multiple packages stay visible and reachable without drilling.

Behaviors:
- All discovered packages default to expanded — matches the user's expectation of seeing everything at once; user can collapse rows they don't care about.
- Search flattens to a list of matching snippets regardless of nesting, each annotated with its package path so the origin is still clear.
- Implicit ancestor packages (e.g. \`a/b/c\` implies \`a\` and \`a/b\`) are materialized so deeply nested snippets aren't orphaned if an intermediate package isn't listed explicitly.
- Depth-based left padding + chevron rotation mirror \`SftpPaneTreeView\` affordances.

## Test plan
- [x] \`tsc --noEmit\` clean on \`ScriptsSidePanel.tsx\`.
- [x] \`eslint components/ScriptsSidePanel.tsx\` clean.
- [x] Manual: open the terminal scripts side panel with multiple nested packages; confirm chevron expand/collapse works per row, snippet click runs in the focused terminal, context menu still offers Edit/Delete.
- [x] Manual: search; confirm matching snippets appear flattened with their package path visible as subtitle, and that clearing search restores the tree (with previous expand state).
- [x] Manual: orphan (no-package) snippets render at top level.

### DIFF
--- a/components/ScriptsSidePanel.tsx
+++ b/components/ScriptsSidePanel.tsx
@@ -8,7 +8,7 @@
  */
 
 import { ChevronRight, Edit2, FileCode, Package, Plus, Search, Trash2, Zap } from 'lucide-react';
-import React, { memo, useCallback, useEffect, useMemo, useState } from 'react';
+import React, { memo, useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { useI18n } from '../application/i18n/I18nProvider';
 import { cn } from '../lib/utils';
 import { Snippet } from '../types';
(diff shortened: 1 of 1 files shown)


binaricat:
@codex review

PR #810, opened 2026-04-22
fix(ssh): include legacy HMACs for very old servers (closes #807)

### LINKED ISSUE #807: Freebsd  6.1 无法连接
报错：
Error invoking remote method 'netcatty:start': Error: Handshake failed: no matching key exchange algorithm

当前版本：
1.0.96

binaricat:
Closes #807.

## Root cause

`buildAlgorithms()` in `electron/bridges/sshBridge.cjs` handles the "allow legacy algorithms" toggle by appending legacy entries to **kex**, **cipher**, and **serverHostKey** — but never touches `algorithms.hmac` at all. So even with the toggle on, the ssh2 library's default (modern) HMAC list applied, which doesn't include `hmac-sha1` / `hmac-md5`.

Very old OpenSSH builds (FreeBSD 6.1, released 2006 — per the issue) can only speak `hmac-sha1` or `hmac-md5`. With the toggle on, kex and host-key negotiation succeeded but MAC negotiation settled on something the server couldn't actually produce. The wrong exchange-hash MAC then broke host-key signature verification, surfacing as:

> \`Handshake failed: signature verification failed\`

The message looks like a host-key algorithm problem, which is why the user's initial workaround (enabling `ssh-dss`/`ssh-rsa`) didn't help — the real mismatch was on the MAC layer.

## Fix

Add an explicit `algorithms.hmac` list to the legacy branch that keeps modern MACs in front and appends the two legacy ones:

\`\`\`js
algorithms.hmac = [
  'hmac-sha2-256-etm@openssh.com', 'hmac-sha2-512-etm@openssh.com',
  'hmac-sha2-256', 'hmac-sha2-512',
  'hmac-sha1', 'hmac-md5',
];
\`\`\`

Modern servers still prefer SHA-2 (they appear first and server picks the first match from its own list). Only servers that literally cannot do SHA-2 will fall back to SHA-1/MD5 — the exact behavior the toggle promises.

## Test plan

- [x] `node -c electron/bridges/sshBridge.cjs` — syntax OK
- Needs verification from the issue reporter against FreeBSD 6.1 — the fix matches their local ssh command parameters (`MACs=+hmac-sha1,hmac-md5`).
- No regression risk on modern servers: HMAC list is additive, SHA-2 entries come first, so servers that can do SHA-2 still will.

### DIFF
--- a/electron/bridges/sshBridge.cjs
+++ b/electron/bridges/sshBridge.cjs
@@ -7,6 +7,7 @@ const net = require("node:net");
 const fs = require("node:fs");
 const path = require("node:path");
 const os = require("node:os");
+const crypto = require("node:crypto");
 const { exec } = require("node:child_process");
 const { Client: SSHClient, utils: sshUtils } = require("ssh2");
 const { NetcattyAgent } = require("./netcattyAgent.cjs");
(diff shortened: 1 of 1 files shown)


binaricat:
@codex review

binaricat:
@codex review

PR #812, opened 2026-04-22
feat(terminal): quick encoding switch for telnet & serial (closes #804)

### LINKED ISSUE #804: [Feature Request]所有协议都支持快速切换终端编码
现状：目前只有在ssh协议时才能在终端快速切换编码

Image

期望：在telnet和串口协议下也支持快速切换终端编码

binaricat:
Closes #804.

## Motivation

The toolbar's UTF-8 / GB18030 quick switch only appeared for SSH sessions — telnet and serial were gated out. Their backend decoder was also hardcoded at session start via Node's \`StringDecoder\`, which doesn't know GB18030, so a serial console or legacy telnet host emitting GBK had no path to being rendered correctly.

### Main (\`electron/bridges/terminalBridge.cjs\`)
- Swap \`StringDecoder\` for \`iconv-lite\` on the **telnet** and **serial** paths — iconv actually understands GB18030. Local PTY and mosh stay on \`StringDecoder\` (local follows the OS locale, mosh frames its own UTF-8; neither needs runtime swaps).
- Store the decoder via a mutable \`decoderRef\` on the session so the \`onData\` closures stay untouched while a new IPC can swap in a fresh decoder mid-session.
- New \`normalizeTerminalEncoding\` resolves user-facing charset names (\`utf-8\` / \`gbk\` / \`gb2312\` / \`gb18030\`) to iconv identifiers.
- Register \`netcatty:terminal:setEncoding\`. Mirrors the new encoding onto \`session.serialEncoding\` too, so \`aiBridge\` / \`mcpServerBridge\` exec paths that still read the legacy field pick it up.

### Preload (\`electron/preload.cjs\`)
- \`setSessionEncoding\` tries the existing SSH handler first, then falls through to the new terminal handler when the SSH handler reports \`ok: false\` (non-SSH sessions lack \`session.stream\`, so it already short-circuits for them). Single preload method, one extra IPC round-trip only for telnet/serial and only on explicit user click.

### Renderer
- \`TerminalToolbar\`: drop the \`isSSHSession\` gate. New \`encodingSwitchSupported\` = not local, not mosh, not localhost-PTY.
- \`Terminal.tsx\` \`onSessionAttached\`: sync initial encoding for every protocol that supports it (same gate as the toolbar), not just SSH.

## Test plan

- [x] \`node -c\` on \`terminalBridge.cjs\` and \`preload.cjs\` — syntax clean.
- [x] \`npx tsc --noEmit\` — error count unchanged vs main (24 baseline, 24 with this change).
- [x] \`npx eslint components/Terminal.tsx components/terminal/TerminalToolbar.tsx\` — clean.
- Manual (needs verification):
  - [x] Telnet session to a GBK server → menu shows both options, switching to GB18030 renders Chinese output correctly.
  - [x] Serial session to a device emitting GBK (or UTF-8) → same flip works without disconnecting.
  - [x] SSH session — unchanged behavior, menu still present, switch still works.
  - [x] Local PTY / mosh — menu hidden as before.
  - [x] Switch encoding mid-stream → partial-byte state from the old decoder doesn't leak (fresh decoder on every swap).

### DIFF
--- a/components/Terminal.tsx
+++ b/components/Terminal.tsx
@@ -748,7 +748,12 @@ const TerminalComponent: React.FC<TerminalProps> = ({
       // like latin1 / shift_jis that the UI's two-value state can't
       // represent — so syncing them here would clobber that config with
       // whichever of the two the toolbar menu happens to hold.
-      const isSSH = host.protocol !== 'local' && host.protocol !== 'serial' && host.protocol !== 'telnet' && host.protocol !== 'mosh' && !host.moshEnabled && !host.id?.startsWith('local-') && !host.id?.startsWith('serial-') && host.hostname !== 'localhost';
+      // Hostname intentionally isn't part of the gate (matches the
+       // TerminalToolbar encoding menu) — SSH pointed at localhost is
+       // still a real SSH session whose backend starts in utf-8, and
+       // skipping the sync would leave it mismatched with the UI state
+       // the user already picked before reconnecting.
+      const isSSH = host.protocol !== 'local' && host.protocol !== 'serial' && host.protocol !== 'telnet' && host.protocol !== 'mosh' && !host.moshEnabled && !host.id?.startsWith('local-') && !host.id?.startsWith('serial-');
       if (isSSH) {
         setSessionEncoding(id, terminalEncodingRef.current);
       }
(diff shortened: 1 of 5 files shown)


binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

PR #814, opened 2026-04-22
fix(autocomplete): stop prepending theme cwd ("~ ") to completed commands (closes #806)

### LINKED ISSUE #806: Bug: Tab 自动补全产生多余字符串
当netcatty 连接远程服务器后，服务器用的zsh 的robbyrussell主题，prompt前面有个 "~" 符号。

我先运行了sudo id，触发了密码输入交互。然后我在console 随便输入一部分命令，按tab补全，会自动在我的命令前面追加这个 "~ "，导致我的命令被修改。如果我不使用sudo就不会触发。

Image

binaricat:
Closes #806.

## The bug

Reporter connects to a zsh host running the oh-my-zsh **robbyrussell** theme (prompt: `➜  ~ `), runs `sudo id`, enters the password, then types a partial command and hits Tab. The accepted completion shows up on the command line with a phantom `~ ` prepended.

## Root cause

`findPromptBoundary` stops at the **first** `PROMPT_CHAR + space` it finds. Themes that put cwd/branch/VCS info after the prompt char trip it — `"➜  ~ sudo id"` parses as `promptText="➜ "` + `userInput="~ sudo id"`. That junk then rides through the rest of the autocomplete pipeline:

1. `recordCommand` logs entries like `"~ sudo id"` into history.
2. `fuzzyQueryHistory` later returns those polluted entries as suggestions.
3. `insertSuggestion` compares `suggestion.text` (`"~ ls"`) against the detector's `userInput` (`"~ lo"`), falls into the Ctrl-U + rewrite path, and the phantom `~ ` lands on the real command line.

Sudo is a red herring — it's what seeded the bad history fast enough for fuzzy matches to start winning. Without sudo the popup stays empty, so the rewrite path never fires and the user doesn't see the bug.

## Fix

Track what the user actually typed in an independent keystroke buffer (`typedInputBufferRef`) inside the autocomplete hook:

- Append every printable char / paste chunk.
- Pop on backspace, word-kill on `Ctrl+W`.
- Clear on Enter, `Ctrl+C`, `Ctrl+U`, and any escape sequence / unhandled control char (cursor moves we can't follow → invalidate).

New helper `reconcilePromptWithTypedInput(prompt, typed)` inspects the detector's output: if `userInput` ends with the typed buffer and is longer, the parser over-captured — move the excess back to `promptText` so `userInput` matches what was actually typed. Applied at every `detectPrompt` call site (`fetchSuggestions`, the stale-result recheck, `insertSuggestion`).

For Enter-record the typed buffer wins outright when present, but only after a live `detectPrompt` confirms we're at a shell prompt — otherwise a password-entry Enter would log the password as a command. `insertSuggestion` / ghost-text accept update the typed buffer to the accepted text so a subsequent Enter records the right command.

## Test plan

- [x] `npx tsc --noEmit` — error count unchanged (baseline 24).
- [x] `npx eslint` on both touched files — clean.
- Manual (needs verification):
  - [ ] zsh + robbyrussell: type `lo` + Tab with `ls` in history → result is `ls`, **no** `~ ` prepended. History records `ls`, not `~ ls`.
  - [ ] Same host: run `sudo id`, enter password, then type + Tab → completion clean.
  - [ ] Plain bash `user@host:~$ ` prompt — unchanged behavior (detector was already correct here, reconciler is a no-op).
  - [ ] Password entry Enter — does **not** log the password as a command (live `detectPrompt` still returns `isAtPrompt=false` on the `[sudo] password for user: ` line).
  - [ ] Use Ctrl-A / Ctrl-E to move the cursor, then Tab — buffer is invalidated, reconciler defers to the detector (no worse than current behavior).

### DIFF
--- a/components/terminal/autocomplete/promptDetector.ts
+++ b/components/terminal/autocomplete/promptDetector.ts
@@ -206,6 +206,46 @@ function findPromptBoundary(lineText: string): number {
 }
 
 /**
+ * Reconcile a buffer-parsed prompt with the user's own keystroke history.
+ *
+ * findPromptBoundary stops at the first `PROMPT_CHAR + space` it sees, so
+ * themes that render additional content after the prompt char — e.g.
+ * oh-my-zsh's robbyrussell prints "➜  ~ " where `~` is the cwd — get
+ * parsed as prompt="➜ " + userInput="~ lo". Every consumer downstream
+ * (history recording, suggestion matching, insertion) then treats the
+ * theme's cwd marker as part of the user's command, which pollutes
+ * history with entries like "~ sudo id" and makes Tab insertions prepend
+ * a phantom "~ " to the typed command (issue #806).
+ *
+ * Whenever we have an independent record of what the user actually typed
+ * since the last Enter (keystroke buffer), we can detect this case: the
+ * real input is always a suffix of the over-captured userInput. When it
+ * is, reattribute the leading garbage back to promptText so the rest of
+ * the pipeline sees the clean split.
+ */
+export function reconcilePromptWithTypedInput(
+  prompt: PromptDetectionResult,
+  typedInput: string,
+): PromptDetectionResult {
+  if (!prompt.isAtPrompt) return prompt;
+  if (!typedInput) return prompt;
+  if (prompt.userInput === typedInput) return prompt;
+  if (
+    prompt.userInput.length > typedInput.length &&
+    prompt.userInput.endsWith(typedInput)
+  ) {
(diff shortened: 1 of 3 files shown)


binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

PR #815, opened 2026-04-22
fix(autocomplete): sync ghost text to live input on every keystroke

binaricat:
## Problem

Ghost text has a visible \"jitter\" when typing fast. The ghost is computed at fetch time and stored inside \`GhostTextAddon\`, so between a keystroke and the next debounced \`fetchSuggestions\` (~100ms) the on-screen line has advanced but the ghost tail is still the pre-update one. If the user presses → during that window, the old tail is pasted on top of the new char and the command is corrupted:

- Type \`do\` → ghost shows \`cker ls\` (suggesting \`docker ls\`).
- Type \`c\` → ghost still shows \`cker ls\` for ~debounce-ms.
- Press → immediately → writes \`cker ls\` after \`doc\` → line becomes \`doccker ls\`.

## Fix

Two layers:

1. **\`GhostTextAddon.adjustToInput(newInput)\`** — re-render the ghost against a fresh input synchronously, without waiting for the next fetch. If the current suggestion still prefix-matches, shrink / grow the tail. If not, hide (a later fetch will replace it). Called from \`handleInput\` after every buffer mutation (printable / backspace / Ctrl-W / paste tail) when the buffer is reliable.

2. **Defensive recompute at the → and Ctrl-→ accept paths** — instead of trusting \`ghost.getGhostText()\` (which uses the \`show()\`-time input), compute the tail against the live typed buffer. If the suggestion no longer prefix-matches the live buffer, hide without writing. Ctrl-→ additionally resyncs \`ghost.show()\` to the live buffer before calling \`getNextWord\`.

The alignment-gating story introduced in #814 is preserved: when the buffer is unreliable (post ↑ / cursor-move), both the adjust and accept paths skip so the ghost doesn't lie about a line we're not tracking.

## Test plan

- [x] \`npx tsc --noEmit\` — baseline unchanged (24).
- [x] \`npx eslint\` on touched files — clean.
- Manual:
  - [ ] Type \`do\` → see \`cker ls\` ghost → type \`c\` → ghost instantly becomes \`ker ls\` (no visible lag).
  - [ ] Same scenario, press → immediately after typing \`c\` → line becomes \`docker ls\`, not \`doccker ls\`.
  - [ ] Type \`dox\` (no matching suggestion) → ghost disappears on the \`x\` keystroke instead of waiting for the debounce.
  - [ ] Ctrl-→ next-word accept: ghost state stays consistent after each word.
  - [ ] Backspace restores a longer ghost tail when still prefix-matching.

### DIFF
--- a/components/terminal/GhostSuggestionPolicy.test.ts
+++ b/components/terminal/GhostSuggestionPolicy.test.ts
@@ -1,0 +1,22 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+
+import { decideGhostSuggestion } from "./autocomplete/ghostSuggestionPolicy.ts";
+
+test("keeps the active ghost suggestion while input still fits it", () => {
+  const decision = decideGhostSuggestion("docker ps -a", "doc", "docker compose ls");
+
+  assert.deepEqual(decision, { type: "keep" });
+});
+
+test("switches to a new suggestion once the active one no longer matches", () => {
+  const decision = decideGhostSuggestion("docker ps -a", "dog", "dogstatsd");
+
+  assert.deepEqual(decision, { type: "show", suggestion: "dogstatsd" });
+});
+
+test("hides the ghost when neither the active nor next suggestion matches", () => {
+  const decision = decideGhostSuggestion("docker ps -a", "dog", null);
+
+  assert.deepEqual(decision, { type: "hide" });
+});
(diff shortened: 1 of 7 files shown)


binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

PR #893, opened 2026-05-04
Fix Linux/macOS mosh terminfo lookup (#890)

### LINKED ISSUE #890: Bug: 在x64 Linux平台下最新版本netcatty无法使用mosh
## Describe the problem
和#888一样的问题，我在ubuntu 26 04上测试mosh报错

Image

## Steps to reproduce
1.使用1.1.0版本测试mosh协议连接正常
2.使用Netcatty-1.1.1-linux-amd64.deb包升级软件版本。
3.重新测试报错Error: Terminfo database could not be found.[mosh is exiting.]

## Expected behavior
开启mosh连接后应该正常连接
## Actual behavior
无法正常链接
## Environment
- App: netcatty 1.1.1
- Platform: linux
- UA: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) netcatty/1.1.1 Chrome/144.0.7559.236 Electron/40.8.5 Safari/537.36

binaricat:
## Summary
- Fixes #890. After 1.1.1 switched from the system `mosh` to a bundled, statically-linked `mosh-client`, the embedded ncurses' compiled-in default terminfo path points at the build-time `--prefix=$WORK/prefix` — a temp dir that no longer exists on the user's machine. On a clean Ubuntu 26.04 install `setupterm("xterm-256color")` then fails with `Error: Terminfo database could not be found.`
- The Windows fix in #889 only set `TERMINFO`/`TERMINFO_DIRS` on `win32`. This PR extends `addBundledMoshTerminfoEnv` to POSIX: on Linux it points `TERMINFO_DIRS` at the standard distro paths (`/etc/terminfo:/lib/terminfo:/usr/share/terminfo:/usr/lib/terminfo`); on macOS it adds Homebrew / MacPorts locations.
- A sibling `terminfo/` directory next to the binary (future-proofing for a Linux bundle) is prepended when present, and any caller-supplied `TERMINFO_DIRS` is respected.

## Why mac wasn't already broken
macOS users tend to have one of: a system `/usr/share/terminfo`, a `~/.terminfo`, or a shell-exported `TERMINFO_DIRS` that ncurses' first three search steps hit before reaching the broken compiled-in fallback. Linux users on a fresh `.deb` install don't have those, so the fallback fires immediately. After this change all three POSIX targets behave deterministically.

## Test plan
- [x] `node --test electron/bridges/terminalBridge.bareMoshClient.test.cjs` (16/16, including 4 new cases for Linux defaults, bundled-dir prepend, caller override, and Darwin paths)
- [x] `node --test electron/bridges/terminalBridge.bundledMosh.test.cjs electron/bridges/terminalBridge.moshHandshakeSession.test.cjs electron/bridges/moshHandshake.test.cjs` (39/39)
- [ ] Manual: install the resulting `.deb` on Ubuntu 26.04 (and a stripped container without `ncurses-base` if available), connect via mosh, verify the session attaches without the terminfo error
- [ ] Manual: macOS smoke test that bundled mosh still connects (no regression on the already-working path)

### DIFF
--- a/electron/bridges/terminalBridge.bareMoshClient.test.cjs
+++ b/electron/bridges/terminalBridge.bareMoshClient.test.cjs
@@ -125,6 +125,61 @@ test("Windows dev mosh-client updates the PATH key used by child process env", (
   assert.equal(Object.prototype.hasOwnProperty.call(env, "Path"), false);
 });
 
+test("Linux mosh-client points ncurses at standard distro terminfo paths", () => {
+  const tmp = makeTmp();
+  const client = path.join(tmp, "resources", "mosh", "linux-x64", "mosh-client");
+  writeExecutable(client);
+
+  const env = {};
+  addBundledMoshTerminfoEnv(env, client, { platform: "linux" });
+
+  assert.equal(env.TERMINFO, undefined);
+  const dirs = env.TERMINFO_DIRS.split(":");
+  assert.ok(dirs.includes("/etc/terminfo"));
+  assert.ok(dirs.includes("/lib/terminfo"));
+  assert.ok(dirs.includes("/usr/share/terminfo"));
+});
+
+test("Linux mosh-client prepends a sibling bundled terminfo dir when present", () => {
+  const tmp = makeTmp();
+  const client = path.join(tmp, "resources", "mosh", "linux-x64", "mosh-client");
+  const terminfo = path.join(tmp, "resources", "mosh", "linux-x64", "terminfo");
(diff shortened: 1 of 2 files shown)


binaricat:
Superseded by #894 — that PR ships a private terminfo database inside the mosh binary release so mosh-client is self-sufficient on every platform, instead of relying on the host having `ncurses-base`. Closing this in favor of the deeper fix.

PR #943, opened 2026-05-11
Fix AI export menu theme colors

DeepFal:
Summary: Replace the AI chat export dropdown's hard-coded dark colors with theme-aware popover/accent tokens so it stays readable in light mode and dark mode. Testing: git diff --check. Full lint/test not run because node_modules is not installed in this workspace.


DeepFal:
Closing temporarily to install dependencies locally and re-run validation before reopening.

PR #963, opened 2026-05-12
Fix #919: harden built-in Telnet handshake for legacy gear

### LINKED ISSUE #919: Connects fine but keystrokes not recognized
Hello,

I have no issues with most of my telnet connections however, on a particular Switch, the connection is successful but cant get passed "Press any key to continue" with a blinking cursor.    I can use the Local Terminal just fine and no issues.  Its just with the built-in telnet feature that it doesn't respond. Anyone else ? 

Thanks

binaricat:
## Summary

Addresses #919 — the built-in Telnet client got stuck at "Press any key to continue" on an HP ProCurve 2610, and after a subsequent update the same session started showing random-looking characters in the username field. Three independent correctness gaps in the old protocol layer, all fixed in one PR.

## Root causes

1. **Negotiation parser was stateless per chunk.** [`handleTelnetNegotiation`](electron/bridges/terminalBridge.cjs) ran on each `data` event in isolation. An IAC sequence split across TCP frames either dropped the lone IAC (lost command), or — worse — for `IAC SB ... IAC SE` blocks whose terminator landed in the next frame, fell through to "skip IAC SB and treat the rest as data". That spilled the subnegotiation payload (TERMINAL-TYPE strings, environment data) into the terminal display as garbage. This is the most likely source of the "Username: z Wz Wz W3" symptom in the issue thread.

2. **Client was purely reactive.** It only ever responded to options the server raised. Legacy switch / router firmware often waits for the client to actively commit to `SUPPRESS-GO-AHEAD` / `TERMINAL-TYPE` / `NAWS` before advancing past the banner — so the connection silently hung forever at "Press any key to continue".

3. **Outbound user input was never IAC-escaped.** Any 0xFF byte the user pastes (or that an alternate input encoding emits) was read by the peer as the start of a command, eating the following byte. UTF-8 keyboard typing does not produce 0xFF, but paste of binary content and some non-ASCII encodings do.

### New `electron/bridges/telnetProtocol.cjs`

Pure module owning RFC 854 framing, separable from the bridge so it can be unit-tested without a socket.

- `createTelnetParser({ onData, onCommand, onSubnegotiation })` — a stateful machine that buffers any partial command (lone IAC, IAC + verb without option, unterminated SB) across feeds and replays it once the remaining bytes arrive. Emits clean stream bytes (with `IAC IAC` already unescaped), option commands (`WILL/WONT/DO/DONT`), and complete subnegotiations (with payload `IAC IAC` unescaped) through callbacks. Ordering preserved: data emitted before a command is delivered before that command's callback fires.
- `escapeIacForWire(buf)` — doubles 0xFF bytes on the way out, with a cheap fast-path that returns the input untouched when no 0xFF is present.

### `terminalBridge.cjs` rewiring

- **Lazy protocol activation.** Until the peer sends an IAC byte we leave the data stream alone (raw passthrough). True raw-TCP-on-port-23 services (the long tail of embedded gear) are no longer corrupted by a misread of an incidental 0xFF byte.
- **Active negotiation on activation.** When the peer's first IAC arrives we proactively request `DO SUPPRESS-GO-AHEAD`, `WILL TERMINAL-TYPE`, `WILL NAWS`, and track those in a `requestedOptions` Set. The peer's acknowledgement is consumed silently, so we no longer ping-pong replies on each WILL/WONT/DO/DONT.
- **Uppercase TERMINAL-TYPE.** We advertise `XTERM-256COLOR` (upper-case); legacy boxes that do case-sensitive `termcap` lookups recognise it where the previous `xterm-256color` failed and fell back to dumb terminal.
- **Resize-driven NAWS** only fires after the protocol has actually activated, so a passthrough session is never poisoned by stray IAC bytes.
- **Outbound user input** for telnet sockets goes through `escapeIacForWire` (strings are converted to UTF-8 buffers first), so paste of binary content and non-ASCII encodings round-trip safely.

## Tests

17 new tests in [`electron/bridges/telnetProtocol.test.cjs`](electron/bridges/telnetProtocol.test.cjs):

- Plain data passthrough
- Single-frame option commands and subnegotiations
- `IAC IAC` unescape in both data stream and SB payload
- **Every cross-frame split point**: lone IAC at end-of-chunk, `IAC + DO` without option, mid-SB without `IAC SE` — each correctly buffered and resumed on the next feed
- The specific regression that previously leaked SB payload as data when the SE never arrived
- Two consecutive option commands without losing either
- Callback ordering (data → command → data) matches byte order in the input

All 18 existing telnet auto-login tests still pass, exercising the end-to-end socket → parser → renderer path. Full suite: **825 pass / 0 fail / 3 skipped**.

\`npx tsc --noEmit\` clean on changed files (other pre-existing repo-wide TS errors unrelated to this PR).

## Test plan

- [x] Unit tests cover parser state and IAC escape
- [x] Existing telnet auto-login integration tests still pass
- [ ] Manual: connect to an HP ProCurve 2610 / similar legacy switch, confirm the "Press any key to continue" prompt now advances on Enter and the username field shows typed characters cleanly
- [ ] Manual: connect to a modern Linux telnet server (or [\`busybox telnetd\`]), confirm no regression in normal usage
- [ ] Manual: connect to a raw-TCP-on-23 service (e.g. nc -l 23 with no protocol), confirm data flows raw (no negotiation injected, no garbage)

### DIFF
--- a/electron/bridges/telnetProtocol.cjs
+++ b/electron/bridges/telnetProtocol.cjs
@@ -236,6 +236,142 @@ function unescapeIacFromPayload(buf) {
   return Buffer.from(out);
 }
 
+/**
+ * Build a Telnet negotiation policy machine.
+ *
+ * The machine owns the rules for which options we accept, the
+ * direction-aware acknowledgement tracking, and the wire bytes sent in
+ * response to peer commands. It is intentionally separated from socket I/O
+ * so it can be exercised directly in unit tests.
+ *
+ * `writeCommand(cmd, opt)` is invoked to send `IAC <cmd> <opt>` on the
+ * wire; `writeSubnegotiation(opt, payload)` is invoked to send
+ * `IAC SB <opt> <payload...> IAC SE` (with `payload` already escaped if
+ * needed by the caller); `getWindowSize()` returns `{ cols, rows }` for
+ * the current terminal dimensions; `termType` is the string advertised
+ * for TERMINAL-TYPE subnegotiation (default "XTERM-256COLOR").
+ */
+function createTelnetNegotiator({
+  writeCommand,
+  writeSubnegotiation,
+  getWindowSize,
+  termType = "XTERM-256COLOR",
+} = {}) {
+  const pendingDoRequests = new Set();
+  const pendingWillRequests = new Set();
+
+  const noopWrite = () => {};
+  const cmdSink = typeof writeCommand === "function" ? writeCommand : noopWrite;
+  const sbSink = typeof writeSubnegotiation === "function" ? writeSubnegotiation : noopWrite;
+  const sizeFn = typeof getWindowSize === "function"
+    ? getWindowSize
+    : () => ({ cols: 80, rows: 24 });
+
+  const naws = () => {
+    const { cols, rows } = sizeFn() || {};
+    const safeCols = Number.isFinite(cols) && cols > 0 ? cols : 80;
+    const safeRows = Number.isFinite(rows) && rows > 0 ? rows : 24;
+    const payload = Buffer.from([
+      (safeCols >> 8) & 0xff, safeCols & 0xff,
+      (safeRows >> 8) & 0xff, safeRows & 0xff,
+    ]);
+    sbSink(OPT.NAWS, escapeIacForWire(payload));
+  };
+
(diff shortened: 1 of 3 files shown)


binaricat:
@codex review

PR #1050, opened 2026-05-22
fix #1049: clear WebGL texture atlas to recover from garbled terminal

### LINKED ISSUE #1049: 花屏、乱码
终端用着用着就会出现花屏/乱码。

如果不关闭终端会一直显示花屏/乱码。

重新打开一个新的终端会正常显示。

目前在两台电脑上出现过此情况。

Image
Image
Image
Image
Image

binaricat:
## Problem

Issue #1049: terminal output becomes garbled and **stays** garbled — closing and reopening the *same* terminal does nothing, only opening a brand-new terminal recovers. Reported on **Windows 11** while running heavy AI CLI agents (claude code / opencode / gemini cli) over SSH.

The "only a new terminal fixes it" signature means the corruption lives in the **renderer instance**, not in the data from the server (so it's not an encoding/UTF-8 problem). xterm.js's WebGL renderer caches rasterized glyphs in a **texture atlas**. Two things push that atlas into a corrupted-for-life state:

1. **devicePixelRatio changes** — moving the window between monitors with different DPI, or changing OS display scaling (very common on Windows). The cached glyphs are then drawn at the wrong scale. xterm.js's documented recommendation is to call `clearTextureAtlas()` on DPR change.
2. **Font changes** while a session is live (xterm.js [#3280](https://github.com/xtermjs/xterm.js/issues/3280) — renderer keeps drawing stale glyphs from the old metrics).

Heavy full-screen TUIs (the AI agents) make this much more likely because they hammer the atlas with spinners, box-drawing, emoji, and constant redraws.

## Fix

Clear the texture atlas so glyphs re-rasterize at the correct scale, instead of forcing the user to reopen the terminal. All paths are **no-ops under the DOM renderer**.

- **`rendererDprWatch.ts`** (new, TDD'd): `watchDevicePixelRatio()` registers a `matchMedia('(resolution: Ndppx)')` listener and **re-registers** across each DPI change (a media query only matches the ratio it was created with), with a legacy `addListener`/`removeListener` fallback. Returns a cleanup fn.
- **`createXTermRuntime.ts`**:
  - on devicePixelRatio change → clear atlas + refit (covers the Windows scaling / multi-monitor trigger; `matchMedia` change does not fire a normal resize, so this is needed on top of resize handling);
  - clear atlas on reflow (`term.onResize`);
  - expose `clearTextureAtlas()` on `XTermRuntime`; tear the watcher down on `dispose`.
- **`Terminal.tsx`**: call `clearTextureAtlas()` after the font/appearance effect applies font options.

## Verification

- `node --test` runtime suite: **135 pass / 0 fail** (incl. 5 new `rendererDprWatch` tests, written test-first / red→green).
- `tsc --noEmit`: **40 errors on both this branch and `origin/main`** → 0 new type errors introduced (all pre-existing, in unrelated files).
- `eslint` on all changed files: clean (exit 0).
- `vite build`: exit 0.

## Notes / follow-ups (not in this PR)

- Suggested the reporter switch **Settings → Terminal → Renderer = DOM** as an immediate workaround / confirmation.
- Lower-priority follow-ups from triage left out to keep this focused: a user-facing "repair garble" shortcut, and improving `onContextLoss` to fully fall back/recreate rather than just dispose.

### DIFF
--- a/components/Terminal.tsx
+++ b/components/Terminal.tsx
@@ -1241,6 +1241,11 @@ const TerminalComponent: React.FC<TerminalProps> = ({
         termRef.current.options.ignoreBracketedPasteMode = terminalSettings.disableBracketedPaste ?? false;
       }
 
+      // Changing the font can leave the WebGL renderer drawing stale glyphs from
+      // the old metrics (xterm.js #3280), surfacing as garbled text (issue #1049).
+      // Clear the texture atlas so glyphs re-rasterize with the new font.
+      xtermRuntimeRef.current?.clearTextureAtlas();
+
       if (isVisibleRef.current) {
         setTimeout(() => safeFit({ force: true, requireVisible: true }), 50);
       } else {
--- a/components/terminal/runtime/createXTermRuntime.ts
+++ b/components/terminal/runtime/createXTermRuntime.ts
@@ -43,6 +43,7 @@ import {
 } from "./kittyKeyboardProtocol";
 import { installKittyKeyboardProtocolHandlers } from "./kittyKeyboardRuntime";
 import { installUserCursorPreferenceGuard } from "./cursorPreference";
+import { watchDevicePixelRatio } from "./rendererDprWatch";
 import { handleSerialLineModeInput } from "./serialLineInput";
 import {
   markExpectedTerminalCursorPositionReport,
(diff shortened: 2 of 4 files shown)


binaricat:
@codex review

PR #1107, opened 2026-05-26
Per-agent provider switcher chip in the Catty Agent chat input

binaricat:
## Summary
Catty Agent currently has no model chip in the chat input — \`modelPresets\` is empty for the catty agentId so \`hasModelPicker\` is false, and the provider/model the chat ends up using comes from the global \`activeProviderId\` / \`activeModelId\`. Switching providers means opening Settings → AI → Providers, toggling the active one, and coming back. Closes the gap surfaced in the original feedback on #1101 (problem 2) and partially addresses #986 (per-context model switching) by going per-agent.

- **State**: new \`agentProviderMap\` (\`Record<agentId, providerId>\`) sits next to the existing \`agentModelMap\` with its own storage key + cross-window sync. Both maps are written together when the user picks from the chip; together they form a per-agent override that beats the global active provider/model. ACP agents (Claude/Codex) plumb their provider through the CLI, so the override doesn't apply to them.
- **ChatInput**: new \`providerSwitcher\` prop. When set, the model chip switches from the Cpu glyph to a \`ProviderIconBadge\` + \`providerName · modelId\` label, and the popover renders a flat list — provider icon (size=md) + name + the configured default model as caption (mono if set, italic hint if not) + a Check on the bound row. One provider has one model (its \`defaultModel\`), so a two-level menu would be theatre; one click is enough.
- **List visibility**: the picker lists every configured provider, matching what Settings → AI → Providers shows. The per-provider \`enabled\` toggle is an "active-ish" flag, not a visibility gate; hiding disabled providers silently made everything but the one currently active disappear in early feedback.
- **AIChatSidePanel**: resolves \`effectiveActiveProvider\` / \`effectiveActiveModelId\` for the catty agent from the per-agent map, falling back to global. \`handleSend\` passes those through to \`sendToCattyAgent\` so a provider picked in one Catty session applies everywhere Catty runs.
- **Sizing**: \`ProviderIconBadge\` grew an \`xs\` size (16px badge, 10px glyph) so the chat-input chip doesn't crowd the \`h-6\` toolbar row; the popover bumps to \`md\` (32px badge).
- **i18n**: adds \`ai.chat.noProviderModel\` in en / zh-CN / ru.

For v1, each provider's row shows just its configured \`defaultModel\`. The two-level structure is gone (one provider → one model), but the data flow is general — a future PR can surface multiple known models per provider (cached \`/models\`, manual additions) by adding rows to the popover without revisiting the chip wiring.

Refs #1101, #986. Stacks cleanly on top of #1105.

## Test plan
- [ ] \`npm test\` — 1249 pass locally.
- [ ] \`npm run build\` — clean.
- [ ] Settings → AI → Providers: configure 2+ providers (any \`enabled\` state). Open Catty Agent: chip shows the active provider's icon + name + model.
- [ ] Click the chip: popover lists every configured provider (not just enabled ones), matches the Settings page.
- [ ] Pick a different provider: chip updates immediately; send a message and verify it goes to the new provider.
- [ ] Switch agent to Claude/Codex: chip falls back to the existing model-presets dropdown unchanged.
- [ ] Reload: per-agent provider/model selection persists.


binaricat:
@codex review

Issue #1120, opened 2026-05-27
Autocomplete popup does not disappear in some cases


pyroch:
In some cases autocomplete popup does not disappear after sending command (on video)

https://github.com/user-attachments/assets/9470cf3a-7dbc-41d2-a5e7-de3176dcf140

PR #1121, opened 2026-05-27
(fix) autocomplete popup visible after sending command

pyroch:
Fixes (https://github.com/binaricat/Netcatty/issues/1120) a race where the terminal autocomplete popup could remain visible or reappear after pressing Enter, most noticeably on higher ping hosts.

The issue was caused by pending debounced autocomplete fetches surviving after autocomplete state had already been cleared. If one of those delayed fetches completed after Enter closed the popup, it could apply stale suggestions back into state.

Changes:

Cancel any pending autocomplete debounce timer when clearing autocomplete state.
Reset the debounce timer ref after the timer fires.
Add a final version check inside the React transition state update so stale async completion results cannot reopen the popup.
Validation:

Ran npx eslint components\terminal\autocomplete\useTerminalAutocomplete.ts successfully.

### DIFF
--- a/components/terminal/autocomplete/useTerminalAutocomplete.ts
+++ b/components/terminal/autocomplete/useTerminalAutocomplete.ts
@@ -382,6 +382,10 @@ export function useTerminalAutocomplete(
    * Clear popup/ghost state. Skips re-render if already empty.
    */
   const clearState = useCallback(() => {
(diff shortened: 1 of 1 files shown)


binaricat:
Nice catch — thanks for the fix!

PR #1125, opened 2026-05-27
fix(terminal): probe cwd via SSH_CONNECTION on older OpenSSH (#1123)

### LINKED ISSUE #1123: Bug:定位到终端当前目录无效
无法定位到终端当前目录
版本 1.1.15

https://github.com/user-attachments/assets/e8537c2f-6d01-4c64-a84d-dec76cc1626a

binaricat:
Fixes #1123.

## Summary

- On CentOS 7 / RHEL 7 (OpenSSH 6.6.1) the "go to terminal cwd" feature always returns `$HOME` instead of the user's actual `cd`-ed directory.
- Root cause: the pwd-probe script assumed the user's PTY shell and the probe's exec channel are siblings (same per-connection sshd as `$PPID`). That holds on newer OpenSSH, but old OpenSSH forks a *separate* sshd child per channel — the PTY shell is a *cousin* (same grandparent sshd, different parent), so the sibling lookup returns nothing and we drop into the home-dir fallback.
- Added a second pass that matches by the `SSH_CONNECTION` env var. sshd injects the unique client-port / server-port 4-tuple into every channel of one SSH connection, so processes that share it are exactly the channels of this connection. We scan `/proc/<pid>/environ` (same-uid readable), filter for a shell `comm` with a controlling tty, and feed that PID into the existing `readlink /proc/<pid>/cwd` step.
- The sibling search stays as the fast path; the new `SSH_CONNECTION` lookup only runs when the fast path returns nothing. No change for non-Linux remotes (no `/proc` → unchanged home-dir fallback).

## Test plan

- [ ] CentOS 7 host: connect, `cd /tmp`, click "go to terminal cwd" — SFTP should land in `/tmp`, not `/root`. Try a few other dirs and nested `sudo bash` to confirm the `find_active_shell` step still walks down to the active shell.
- [ ] Modern Ubuntu / Debian host: same flow still works (fast path hits, fallback never runs)
- [ ] Multi-session same host: open two Netcatty terminals to the same server under the same user, each `cd` to a different directory, probe each — they should not cross-contaminate (`SSH_CONNECTION` 4-tuple is per-connection, not per-user)
- [ ] BSD / macOS remote (no `/proc`): unchanged — still falls back to `$HOME`
- [ ] Network device host classified by `shouldProbeSessionCwd` as non-probeable: probe is still skipped upstream, untouched

### DIFF
--- a/electron/bridges/sshBridge.cjs
+++ b/electron/bridges/sshBridge.cjs
@@ -2059,7 +2059,9 @@ async function getSessionPwd(event, payload) {
     }, 5000);
 
     // POSIX sh script that:
-    //   1. Finds the sibling interactive shell under sshd ($PPID).
+    //   1. Finds the user's interactive shell on the same SSH connection
+    //      (sibling under $PPID on newer OpenSSH, cousin reachable via the
+    //      shared SSH_CONNECTION env var on older OpenSSH like CentOS 7).
     //   2. Follows foreground child shells only, which covers bash->fish
     //      without mistaking background shell scripts for the active shell.
     //   3. Reads /proc/<pid>/cwd via readlink.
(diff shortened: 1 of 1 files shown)


binaricat:
@codex review

PR #1163, opened 2026-06-01
feat(vault): connect with the host's default protocol, prompt only for Telnet

binaricat:
Addresses part 1 of #1099 — a Telnet host showing the protocol picker on every connect.

## What changed

Connecting now resolves to a clear default, and the picker only appears when it's genuinely a choice:

| Host setup | Connect (single-click & batch) |
|---|---|
| Telnet **set as default** (`protocol = telnet`) | Telnet — no picker |
| Telnet enabled but **not** default | **picker** (unchanged behavior) |
| Mosh enabled | Mosh — no picker |
| otherwise | SSH |

So the only thing that pops the picker is "Telnet is available but isn't this host's default". SSH-only, SSH+Mosh, and Telnet-default all connect directly.

## UI

The Telnet card in host settings gains a **"Connect with Telnet by default"** switch. Turning it on makes Telnet the host's primary protocol so it connects straight through; off restores SSH as primary while keeping Telnet available (which falls back to the picker).

## Notes / behavior changes

- **SSH+Mosh now connects Mosh directly** instead of prompting every time. If you enabled Mosh, it's used. (To use SSH on such a host, disable Mosh.)
- There's intentionally no "SSH-default while keeping Telnet, don't ask" — keeping Telnet available but not default means you get the picker (Telnet vs SSH is a deliberate, different-protocol/credentials choice). Set one as default to skip it.
- Batch-connecting a "Telnet-enabled-but-not-default" host can't show the picker, so it uses the host's default (SSH/Mosh), not Telnet.
- The earlier draft of this PR used a `skipProtocolPrompt` flag + a "remember" checkbox in the picker; that was dropped in favour of these simpler rules (which also sidesteps the Mosh-clearing edge Codex flagged).
- Parts 2 (Telnet password reveal) and 3 (copy from read-only logs) of the issue are separate, smaller asks — not included here.

## Verification

`eslint` clean on the changed files; `tsc --noEmit` adds no new errors (pre-existing project-wide errors are unrelated). No DOM test harness in the repo, so worth a manual check:
1. Host with SSH + Telnet → settings → toggle "Connect with Telnet by default" → connect goes straight to Telnet.
2. Host with SSH + Mosh → connect → goes straight to Mosh (no picker).
3. Host with SSH + Telnet, toggle off → connect → picker appears as before.

### DIFF
--- a/application/app/AppHandlers.ts
+++ b/application/app/AppHandlers.ts
@@ -736,17 +736,11 @@ export function handleTerminalDataCaptureImpl(getCtx: AppContextGetter, sessionI
 export function hasMultipleProtocolsImpl(getCtx: AppContextGetter, host: Host) {
   const { resolveEffectiveHost } = getCtx();
 {
+    // Gates the protocol picker (legacy name kept for its existing wiring).
+    // Only prompt when Telnet is available but isn't the host's default protocol;
+    // SSH-only, SSH+Mosh and Telnet-default all connect directly.
     const effective = resolveEffectiveHost(host);
-    let count = 0;
-    // SSH is always available as base protocol (unless explicitly set to something else)
-    if (effective.protocol === 'ssh' || !effective.protocol) count++;
-    // Mosh adds another option
-    if (effective.moshEnabled) count++;
-    // Telnet adds another option
-    if (effective.telnetEnabled) count++;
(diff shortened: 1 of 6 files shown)


binaricat:
@codex review

PR #1282, opened 2026-06-06
fix(terminal): simplify sudo password autofill (#1281)

### LINKED ISSUE #1281: Bug: sudo自动代填密码存在的bug
## Describe the problem
1、在ubuntu 2604 上测试不生效无论是sudo -s、sudo -i  sudo [命令] 形式都不带填，应该是没适配2604的sudo提示是[sudo: [sudo] password for {user}: ] Password:
2、在麒麟v11上测试  sudo -i 和sudo -s 正常，sudo [命令] 的代填时好时坏，目前测试，手动输入命令就不能代填，使用提示的补全命令就行。并且即使代填后在终端的光标会跑到Linux命令行提示符的下面一行去（这个感觉不限系统）。

https://github.com/user-attachments/assets/ee325145-cb0c-4cfa-a5b6-4c72485e56d6

https://github.com/user-attachments/assets/91ec7771-0489-48ec-92e0-b928cb550139

## Steps to reproduce
1.准备一台ubuntu 2406的Linux系统
2. netcatty添加主机，开启sudo代填
3. 测试上面的三个命令，可以发现代填无效
4. 再准备一台其他的系统
5. 测试sudo -s和sudo -i 可以发现代填正常但是代填后光标跑到最新的Linux命令行提示符带下方了，退格键可以让它回到正常的位置，但是回车后又跑到下面去了
6. 测试sudo +命令，发现代填失效
## Expected behavior
应该都能自动代填
## Actual behavior
都写情况下代填异常，并且带填后光标位置异常。
## Environment
- App: netcatty 1.1.23
- Platform: win32
- UA: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) netcatty/1.1.23 Chrome/144.0.7559.236 Electron/40.8.5 Safari/537.36

---------------------------------------------------------

更新一下：还发现一个问题sudo 切换root后再次输入 s、su、 sud、sudo 都不在终端打印任何字符但是输入确实完成了，下面 的视频中我其实已经完成输入，但是终端没显示，回车后才显示的

https://github.com/user-attachments/assets/12c68322-10b3-4b8c-9580-1b0cdd2989db

binaricat:
## What

Simplifies sudo password autofill from a command-rewriting scheme to passive prompt observation.

- **Before:** rewrite `sudo X` → `sudo -p '[sudo] password for %p: <random-marker>' X`, wait for the marker in output, fill, then sanitize the marker/echo out (with a Ctrl-U retype).
- **After:** when a `sudo` command is submitted, arm a 10s window; within it, fill the password as soon as a prompt-shaped line appears (ANSI-color tolerant). Input is sent verbatim; output passes through unchanged.

## Why — fixes #1281

| Bug | Root cause | Fix |
|---|---|---|
| Ubuntu never autofilled | PAM overrides `sudo -p`; the injected marker never appears, and fill was gated on that marker | Detection no longer needs a marker; matches a bare `Password:` too |
| Cursor jumps below the prompt | wire command (long) vs displayed command (short) + Ctrl-U retype desync xterm's cursor | command is no longer rewritten |
| Typed cmd fails, autocomplete works | rewrite needed the exact reconstructed command | arming only needs the command to start with `sudo` |

The author's reported Ubuntu prompt `[sudo: [sudo] password for {user}: ] Password:` is covered by a regression test, as is a bare `Password:`. (The nested `[sudo: ...]` wrapper was itself an artifact of the old `-p` injection — it should no longer appear once we stop rewriting.)

## Notes

- Implementation: **322 → 131 lines**.
- Prompt detection strips ANSI color before matching but still refuses SGR-concealed text (anti-spoofing — a remote can't disguise a fake prompt to harvest the password).
- Removed the 3s cooldown: fill-once-then-disarm already prevents repeats, and the cooldown blocked back-to-back sudo commands.
- **Tradeoff:** within the arm window, if you immediately run another program that asks `Password:` (and this sudo didn't prompt due to cached creds), it could fill once. Bounded by the short window + fill-once.

## Test

- `node --test --import tsx components/terminal/runtime/*.test.ts` — green; full suite **1845 pass / 0 fail**.
- Recommend a real-device pass on Ubuntu 24.04/26.04 + Kylin v11 (the issue environments), since sudo+PAM behavior is environment-specific.

### DIFF
--- a/components/Terminal.tsx
+++ b/components/Terminal.tsx
@@ -297,7 +297,8 @@ const TerminalComponent: React.FC<TerminalProps> = ({
     if (!pastedCommand || !shouldRecordShellHistory(pastedCommand[1], termRef.current)) {
       return data;
     }
-    return prepareSudoAutofillInput(data, null, sudoAutofillRef.current);
+    prepareSudoAutofillInput(data, null, sudoAutofillRef.current);
+    return data;
   }, []);
 
   // Terminal autocomplete — onAcceptText writes directly to session (no CustomEvent)
--- a/components/terminal/runtime/createTerminalSessionStarters.et.test.ts
+++ b/components/terminal/runtime/createTerminalSessionStarters.et.test.ts
@@ -5,14 +5,11 @@ import { createTerminalSessionStarters } from "./createTerminalSessionStarters";
 
 const noop = () => undefined;
 
-const prepareSudoPrompt = (
-  autofill: { prepareCommand: (command: string) => string | null } | null,
+const armSudoPrompt = (
+  autofill: { armForCommand: (command: string) => void } | null,
 ): string => {
-  const prepared = autofill?.prepareCommand("sudo whoami");
-  assert.ok(prepared);
-  const prompt = prepared.match(/\s-p '([^']*)'/)?.[1];
-  assert.ok(prompt);
-  return prompt.replace("%p", "alice");
+  autofill?.armForCommand("sudo whoami");
+  return "[sudo] password for alice: ";
 };
 
 const makeBackend = (
@@ -110,7 +107,7 @@ test("startEt enables sudo autofill with the host saved password", async () => {
   };
 
   await createTerminalSessionStarters(ctx as never).startEt(term as never);
-  onData?.(prepareSudoPrompt(sudoAutofillRef.current));
+  onData?.(armSudoPrompt(sudoAutofillRef.current));
 
   assert.deepEqual(sent, ["saved-secret\n"]);
 });
--- a/components/terminal/runtime/createTerminalSessionStarters.mosh.test.ts
+++ b/components/terminal/runtime/createTerminalSessionStarters.mosh.test.ts
@@ -8,14 +8,11 @@ import {
 const noop = () => undefined;
 const ENCRYPTED_CREDENTIAL_PLACEHOLDER = "enc:v1:djEwAAAA";
 
-const prepareSudoPrompt = (
-  autofill: { prepareCommand: (command: string) => string | null } | null,
+const armSudoPrompt = (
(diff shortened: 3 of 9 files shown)


binaricat:
Good catch — this is a real leak, not just the edge case I'd noted. Fixed in dd4d4f95 by splitting prompt detection into two tiers:

- **Explicit sudo prompts** (`[sudo]` tag or `password for <user>`) are filled as before — no other tool prompts this way.
- A **bare prompt** is only filled when the whole line is `Password:` / `密码:`. Prefixed prompts like `Enter password:` (mysql -p) or `user@host's password:` (ssh) are now rejected, so the sudo password can't leak to a child process when sudo's creds are warm or NOPASSWD is set.

Added regression tests for the `sudo mysql -p` and `sudo ssh user@host` cases (both must not fill), alongside the existing `[sudo] password for …:` / bare `Password:` cases.

@codex review

binaricat:
Good catch again — `password for <user>` isn't sudo-specific (psql's `Password for user alice:` is exactly the case you gave). Fixed in bcc389fb: the explicit tier now requires the `[sudo]` tag, and anything without it must be a bare `Password:` / `密码:` line. So `sudo psql -h db -U alice` → `Password for user alice:` no longer fills, alongside the earlier mysql/ssh cases. Regression test added.

@codex review

binaricat:
Fixed in ee9622fd — `armForCommand` now disarms on any non-sudo command, so a pending arm never outlives the next submission. That closes all three P1s from this round (warm-sudo + bare prompt, the `password for <user>` / psql case, and this stale-window case). Merging now.

PR #1283, opened 2026-06-07
feat(terminal): confirm-to-fill sudo password hint (#1281)

binaricat:
## What

Reworks sudo password autofill from **auto-fill** to a **Tabby-style hint + Enter to confirm**.

When a sudo command is armed and a password prompt appears, netcatty shows a dimmed inline hint ("Press Enter to paste sudo password") instead of sending the password. Pressing **Enter** pastes the saved password and submits; any other key dismisses the hint and lets you type it manually.

## Why

Builds on #1282 (which simplified autofill to passive observation). Auto-filling still had a credential-leak class: when sudo's creds are warm / NOPASSWD, sudo stays silent and the program it launches (`mysql -p`, `ssh`, `psql`, …) asks for **its own** password — which auto-fill could satisfy with the saved **sudo** password. Codex flagged three P1 variants of this on #1282.

Requiring explicit confirmation removes that class entirely: nothing is sent unless the user presses Enter at the hint. A detection misfire now only shows a dismissable hint instead of leaking the password.

Because confirmation makes over-matching safe, this also:
- **Relaxes detection** back to a broad "password/密码/口令 + colon" match → covers Ubuntu/PAM bare `Password:`, `[sudo] password for…:`, localized prompts, etc. (the `[sudo]`-only / bare-only narrowing from the codex rounds is no longer needed).
- **Removes the per-host "auto-fill sudo password" toggle** — it's always available when the host has a saved password, no pre-configuration needed.

## How

- `terminalSudoAutofill`: `handleOutput` raises `onHint(true)` instead of writing; `confirmFill()` / `cancelHint()` / `isPromptPending()` added; detection relaxed (SGR-conceal text still rejected as anti-spoofing).
- `GhostTextAddon`: read-only `showHint()` / `hideHint()` overlay, independent of autocomplete suggestions.
- `createXTermRuntime`: while a hint is pending, Enter → `confirmFill`, other keys → `cancelHint`.
- Wiring: `autofill.onHint` → `ctx.onSudoHint` → `sudoHintRef` → `GhostTextAddon`.
- Dropped `terminalSudoAutoFill` from the host model, host-details UI, session-starter gate, and i18n.

## Test

- `terminalSudoAutofill` unit tests rewritten for the hint model (18 cases): hint-not-write, confirmFill writes, cancelHint, relaxed detection, non-sudo disarm, expired arm, conceal rejection.
- Adapted session-starter / attachment tests to the hint model + no toggle.
- Full suite **1846 pass / 0 fail**; lint clean; renderer build OK.
- Manually verified the hint + Enter-to-paste flow on a real Linux host.

Refs #1281.


binaricat:
Both fixed in 820323c9:

- **P1**: `onHint` now returns whether the hint actually rendered. If the overlay is unavailable (autocomplete disabled), `handleOutput` no longer arms a pending confirmation — Enter isn't intercepted and nothing is sent, falling back to manual entry instead of a silent auto-fill. (Fully decoupling the hint overlay from the autocomplete switch would mean changing the ghost-addon lifecycle; this safe fallback covers the security concern for now.)
- **P2**: Escape/Backspace that dismiss the hint are now swallowed (`preventDefault` + consume) so the stray byte never reaches the no-echo password prompt.

binaricat:
@codex review

binaricat:
Fixed in d10eaa4c: `handleOutput` now keeps observing output while a hint is pending and disarms once a new line arrives whose latest line is no longer a password prompt (sudo timed out / failed / returned to the shell). A sudo re-prompt (`Sorry, try again.` + `[sudo] password for…:`) keeps the hint since the latest line is still a prompt. So the pending flag can't outlive the prompt, and a later Enter can't send the password to the shell. Regression tests added for both cases.

PR #1377, opened 2026-06-10
feat(et): support server stats for EternalTerminal sessions

lateautumn233:
- Generalize the Mosh stats companion into reusable connection helpers
- Open a companion SSH connection so the host-info bar works for ET sessions
- Fall back to execOnEtSession for jumped ET sessions without a direct connection
- Forward host-key and algorithm options to the ET backend for companion parity
- Close the ET stats companion on session close, cleanup, and PTY exit


lateautumn233:
image

参考mosh服务器状态的实现，为也et添加对应功能

Issue #1453, opened 2026-06-12
[Bug] 系统监控中进程数和docker的结果不准


lybing911:
### Operating system

Windows

### Netcatty version

1.1.38

### How did you install Netcatty?

GitHub Release (.dmg / .exe / .AppImage / .deb / .rpm / .pacman)

### Affected area

Other

### Can you reproduce it?

Always (100%)

### Steps to reproduce

1. 准备一台openEuler 24.03 (LTS)的系统
2. 使用top或者 ps aux | wc -l查看进程数
3. 和netcatty的系统中的进程数对不上，并且运行状态的进程数也对不上
4. 在系统上安装docker，启动一个测试镜像
5. 在系统中查看docker，提示此主机未检测到docker

docker;

<img width="3072" height="655" alt="Image" src="https://github.com/user-attachments/assets/c63bada3-6ece-440d-9d09-13a66af1c9d6" />

进程数，监控的上限好像是200？

<img width="2323" height="583" alt="Image" src="https://github.com/user-attachments/assets/b7d8d195-bb99-412a-b61e-3106f441852e" />

手动启动10个 run状态的测试进程后

<img width="2382" height="642" alt="Image" src="https://github.com/user-attachments/assets/95f489fd-912e-49f9-90b2-1b82c93557a8" />

### Expected behavior

系统功能正常

### Actual behavior

进程和docker的监控都不准

### Logs / screenshots

Reported from Netcatty Settings (netcatty 1.1.38).

User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) netcatty/1.1.38 Chrome/148.0.7778.218 Electron/42.3.3 Safari/537.36

### Before submitting

- [x] I searched existing issues and did not find a duplicate
- [x] I removed passwords, private keys, and other secrets from this report

binaricat:
这个是重新连接一次依然如此吗，所有的主机都是这样吗

lybing911:
> 这个是重新连接一次依然如此吗，所有的主机都是这样吗

是的，重新找了以太ubuntu 2604的虚拟机测试也是一样的，进程数不准，不管是总的，sleep、run状态的进程数都不准，docker也无法正常监测到

PR #1455, opened 2026-06-12
fix(system): increase process list limit and improve Docker detection for openEuler (#1453)

binaricat:
## Root Cause Analysis

This PR fixes two distinct issues reported in #1453 where system monitoring was inaccurate on openEuler 24.03 LTS.

### Issue 1: Process list capped at 200 entries

**Root cause:** The `PROCESS_LIST_SCRIPT_POSIX` pipeline uses `| head -n 200`, limiting the process list to at most 200 entries. On servers with active workloads (Docker containers, databases, web servers, etc.), the actual process count far exceeds 200, causing the displayed count to mismatch `ps aux | wc -l`.

**Fix:** Increased the limit from 200 to 2000 for both the Linux/SSH (`ps`) and Windows (`PowerShell`) backends. The frontend already uses a virtual list with efficient rendering, so larger lists do not cause performance issues.

### Issue 2: Docker not detected despite being running

**Root cause:** The capability probe script only runs `docker info >/dev/null 2>&1` to detect Docker. On openEuler 24.03 (and potentially other RHEL 9 derivatives), this can fail even when Docker is running, because:
- SSH `conn.exec()` uses a non-interactive, non-login shell which may have different environment/PATH
- Docker socket permissions may differ in non-interactive SSH sessions
- openEuler's Docker configuration may use alternative socket paths

**Fix:** Added a fallback: if `docker info` fails but the Docker socket exists at `/var/run/docker.sock`, Docker is still reported as available. The socket file is created by the Docker daemon at startup and removed on clean shutdown, making it a reliable indicator.

### Changes

```diff
- 'docker info >/dev/null 2>&1 && printf ...'
+ '(docker info >/dev/null 2>&1 || [ -S /var/run/docker.sock ]) && printf ...'

- | head -n 200
+ | head -n 2000

- Select-Object -First 200
+ Select-Object -First 2000
```

### Testing
- `node --check electron/bridges/systemManagerBridge.cjs` ✓ (syntax OK)
- Existing unit tests pass: systemManagerBridge.processes, execConnHealth, tmuxEnv (25 tests)

### DIFF
--- a/application/state/sessionCapabilitiesStore.ts
+++ b/application/state/sessionCapabilitiesStore.ts
@@ -25,6 +25,7 @@ export const sessionCapabilitiesStore = {
     if (!entry) return undefined;
     if (isExpired(entry)) {
       capabilitiesBySessionId.delete(sessionId);
+      notifySession(sessionId);
       return undefined;
     }
     return entry.capabilities;
--- a/application/state/sessionCapabilitiesStore.ts
+++ b/application/state/sessionCapabilitiesStore.ts
@@ -1,9 +1,16 @@
 import type { SessionCapabilities } from '../../domain/systemManager/types';
 
+/** How long cached capability probes remain valid before requiring re-probe. */
+export const CAPABILITIES_TTL_MS = 60_000;
+
 type Listener = () => void;
 
 const capabilitiesBySessionId = new Map<string, SessionCapabilities>();
 const listenersBySessionId = new Map<string, Set<Listener>>();
+
+function isExpired(capabilities: SessionCapabilities): boolean {
+  return Date.now() - capabilities.probedAt > CAPABILITIES_TTL_MS;
+}
 
 function notifySession(sessionId: string) {
   listenersBySessionId.get(sessionId)?.forEach((listener) => listener());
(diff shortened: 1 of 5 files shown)


binaricat:
@codex review

binaricat:
**回复 Codex 评论 #1 (Docker socket `-S` 权限问题)**

This was already addressed in commit 5c53ea67 by switching from `-S` (socket existence) to `-r` (read permission check). The `-S` flag was introduced in commit e080dfb8 (the commit Codex reviewed), which only checked socket existence without verifying permissions. The follow-up commit 5c53ea67 corrected this to `-r`, which checks that the current user can actually read the Docker socket — matching the permission level required to communicate with the Docker daemon.

Codex's second review (on commit 5c53ea67) correctly reflects the updated `-r` version.

**回复 Codex 评论 #2 (进程列表上限)**

Good catch — restored `head -n 2000` (POSIX) and `-First 2000` (Windows) as a safety cap in commit 30b1d740. Added inline comments clarifying that this is a safety guard, not a functional limit: monitored processes still show accurate metrics for the top 2000 processes. Tests pass.

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

PR #1549, opened 2026-06-22
[codex] Fix host keyword highlight enablement

binaricat:
## What changed

- Ensures adding a keyword highlight rule from the current connection panel also enables host keyword highlighting.
- Adds a regression test for adding a new rule when the host already has saved rules but the host-level highlight switch is off.

## Why

A code review found one concrete path matching the issue symptoms for current-connection rules: the rule could be saved successfully while the host-level highlight switch stayed disabled, so the new rule appeared configured but never ran. Existing rules could still appear to work through other enabled/global rules, making the behavior confusing.

I did not find a corresponding code path where global setting rules are saved but dropped. Existing global normalization tests still pass.

Related to #1502.

## Validation

- `node --test --import tsx components/terminal/HostKeywordHighlightPopover.test.ts domain/keywordHighlight.test.ts`
- `npx eslint components/terminal/HostKeywordHighlightPopover.tsx components/terminal/HostKeywordHighlightPopover.test.ts`
- `npm test` (rerun passed after one non-reproducible flaky run)
- `npm run build`

### DIFF
--- a/components/terminal/HostKeywordHighlightPopover.test.ts
+++ b/components/terminal/HostKeywordHighlightPopover.test.ts
@@ -1,0 +1,42 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+
+import type { Host, KeywordHighlightRule } from "../../types.ts";
+import { addHostKeywordHighlightRule } from "./HostKeywordHighlightPopover.tsx";
+
+const baseHost: Host = {
+  id: "host-1",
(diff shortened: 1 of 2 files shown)


binaricat:
Closing this draft for now because the report still needs more detail before we can be confident the fix covers the full issue. I'll follow up on the issue for clarification.

PR #1560, opened 2026-06-22
Add explicit clipboard image upload action

### LINKED ISSUE #1536: [Feature] Add an explicit action to upload clipboard image to remote SSH session
## Problem / pain point

When working inside an SSH terminal, I often need to send a local screenshot to the remote machine and then use the remote file path in CLI tools, Markdown editing, support workflows, or AI coding tools.

Issue #1271 described this workflow through paste, but that design was closed because paste should remain local, instant, and free of hidden network side effects. I agree with that concern. This request is for the same underlying workflow, but through an explicit user action instead of automatic paste handling.

## Proposed solution

Add an explicit terminal action for uploading the current clipboard image to the active SSH session and inserting the resulting remote path into the terminal input.

Suggested UX:

- Add a terminal context-menu item such as `Upload clipboard image and insert path`.
- Only show or enable it for SSH-backed sessions that can open SFTP.
- When clicked, read image data from the local clipboard.
- Convert/save it as PNG internally if needed.
- Upload it through the current session's SFTP support.
- Create the target directory if needed.
- Insert only the remote file path into the terminal input.
- Do not press Enter or auto-submit the command.
- If the clipboard has no image, show a non-blocking message.
- If upload fails, do not insert any path.

This keeps normal paste behavior unchanged and makes the network/file-writing side effect explicit.

Possible implementation approach:

- Reuse Netcatty's existing remote image upload/paste infrastructure from #1408 where possible.
- Keep the explicit action separate from normal `Ctrl+V` / paste behavior.
- Gate the action to active SSH sessions, excluding local, serial, telnet, mosh, and ET sessions unless they are explicitly supported later.
- Use a generated filename with timestamp/random suffix and no spaces.
- Use a managed remote directory, for example `.netcatty-paste-images` under the current remote working directory or another existing Netcatty-managed location.
- Quote or sanitize the inserted remote path so it is safe to paste into a shell command line.

## Alternatives considered

- Automatic Ctrl+V image upload: rejected in #1271 because paste should not silently trigger network I/O, create remote files, or risk uploading private screenshots by accident.
- Drag-and-drop image upload: also a good explicit trigger, but a context-menu action is easier to discover for users who already use terminal right-click actions.
- External script or manual SCP: works, but it breaks the terminal workflow and cannot reliably reuse Netcatty's active SSH/SFTP session.

## Related area

SSH / terminal  
SFTP  
UI / UX

## How important is this to you?

## Before submitting

I searched existing issues and discussions for similar requests.

binaricat:
## Summary
- Move clipboard image upload out of normal terminal paste and into an explicit terminal context-menu action.
- Keep normal text paste and local file-path paste behavior intact.
- Add feedback when the clipboard has no image or the upload cannot complete.

Closes #1536

## Test Plan
- node --test --import tsx components/terminal/clipboardImagePaste.test.ts components/terminal/TerminalContextMenu.test.ts components/terminal/terminalClipboardPaste.test.ts
- npm run lint
- npm test
- npm run build


yh4922:
This feature doesn't work under Claude CLI — right-click doesn't bring up the context menu. I'd like to request a configuration option that makes the default paste button automatically upload the image and paste its path when it detects that the clipboard content is an image.

PR #1650, opened 2026-06-25
fix(terminal): garbled font on Windows cold start (#1647)

### LINKED ISSUE #1647: [Bug] Windows 启动后终端字体显示异常
### Operating system

Windows

### Netcatty version

1.1.42

### Affected area

SSH connection / terminal

### Can you reproduce it?

Always (100%)

### Steps to reproduce

1. 关闭Netcatty
2. 打开Netcatty
3. 连接服务器，查看文件夹下的文件 ls
4. 终端字体显示异常
5. 调整窗口大小后字体恢复正常

### Expected behavior

打开Netcatty后字体正常显示

### Actual behavior

异常字体

Image

调整窗口大小后

Image

### Logs / screenshots

Reported from Netcatty Settings (netcatty 1.1.42).

User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) netcatty/1.1.42 Chrome/148.0.7778.218 Electron/42.3.3 Safari/537.36

### Before submitting

binaricat:
## Background

Fixes #1647: on Windows, the first connection after a cold start renders garbled/misaligned ASCII (e.g. `ls` output); resizing the window restores it.

## Root cause

xterm measures the character cell grid once at `open()`. The default terminal font was `menlo` (platform-independent), which isn't installed on Windows/Linux, so the primary Latin glyphs fell through to the `"JetBrains Mono"` Latin fallback hard-wired into `composeFontFamilyStack` — a bundled `@fontsource` webfont with `font-display: swap`. It loads asynchronously and swaps in *after* xterm has measured the grid, so the glyph advance width no longer matches the locked cell grid → ASCII misaligns. A manual resize triggers `fitAddon.fit()`, which remeasures with the webfont loaded → it recovers. macOS is unaffected because Menlo is a system font (synchronously available, no swap).

The existing `waitForFonts` didn't catch it: it only explicitly preloaded the Nerd Font, not the actual primary font carrying ASCII, and it relied on a single `document.fonts.ready` resolution that can fire before the webfont is even requested — with no remeasure afterward.

## Changes

- **Platform-neutral `auto` font default.** The stored default becomes an `auto` sentinel, resolved to a locally-installed font per OS at render time (win32 → Consolas, linux → DejaVu Sans Mono, darwin → Menlo). `auto` is persisted/synced as-is, so it never leaks one OS's font to another device and never reintroduces #1647 on the receiving machine. Deprecated font ids also migrate to `auto`.
- **General font-ready recovery** (`useTerminalEffects`): explicitly preload the bundled webfonts the terminal actually renders with (JetBrains Mono / Sarasa Mono SC) before remeasuring, and add a `document.fonts` `loadingdone` listener that clears the texture atlas, remeasures, refits and resizes whenever any font finishes loading. Recovery no longer depends on the single `document.fonts.ready` resolution and also covers user-chosen webfonts.

The default change fixes the out-of-the-box Windows case; the font-ready recovery is the general fix.

## Testing

- `node --test --import tsx infrastructure/config/fonts.test.ts lib/fontAvailability.test.ts application/state/settingsStateDefaults.test.ts` — new unit tests for `getDefaultTerminalFontIdForPlatform`, `detectFontPlatform`, `resolveTerminalFontFamilyId`, `isSameResolvedTerminalFont`, `bundledFamiliesInStack`, and deprecated-id migration. All passing.
- `npm run lint` — clean.
- `npm run build` — passing.

Reviewed locally with Codex (multiple adversarial passes); findings on default persistence/sync and side-panel no-op writes were addressed.

### DIFF
--- a/application/state/settingsStateDefaults.test.ts
+++ b/application/state/settingsStateDefaults.test.ts
@@ -7,8 +7,43 @@ import {
   getHslTokenRelativeLuminance,
   resolveReadableForegroundForHsl,
   resolveThemeAccentForeground,
+  migrateIncomingTerminalFontId,
 } from "./settingsStateDefaults.ts";
+import { TERMINAL_FONT_AUTO } from "../../infrastructure/config/fonts.ts";
+import { STORAGE_KEY_TERM_FONT_FAMILY } from "../../infrastructure/config/storageKeys.ts";
 import type { UiThemeTokens } from "../../infrastructure/config/uiThemes.ts";
+
+function installMemoryLocalStorage(): Map<string, string> {
+  const store = new Map<string, string>();
+  (globalThis as { localStorage?: unknown }).localStorage = {
+    getItem: (k: string) => (store.has(k) ? store.get(k)! : null),
+    setItem: (k: string, v: string) => void store.set(k, v),
+    removeItem: (k: string) => void store.delete(k),
+    clear: () => store.clear(),
+  };
+  return store;
+}
+
+test("migrateIncomingTerminalFontId rewrites deprecated ids to the auto sentinel, not menlo", () => {
+  // menlo would put Windows/Linux upgrade users back into the #1647 path,
+  // and a concrete id would leak across OSes via sync; auto resolves per device.
+  const store = installMemoryLocalStorage();
+  try {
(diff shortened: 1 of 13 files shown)


binaricat:
@codex review

PR #1670, opened 2026-06-25
Add latency to terminal network stats

StartHex:
## Summary
- measure SSH server stats round-trip latency alongside network speed
- display latency in the terminal status bar and network details popover
- add localized latency labels and cover the stats response in tests

## Tests
- node --test electron/bridges/sshBridge/sessionOps.serverStats.test.cjs electron/bridges/sshBridge.sessionOps.et.test.cjs


binaricat (approved):
Checked the current diff, confirmed the earlier type-contract comment is already addressed, and ran the related server-stats test plus build/lint locally. Looks good to merge.

PR #1793, opened 2026-06-29
Fix remote autocomplete probe on blank path arguments

### LINKED ISSUE #1193: Bug: 连接公司堡垒机，输入cd,cat 这些常用命令加上空格就会断开
## 问题1：打开这个界面会自动输入一位，布局也有点错位

Image

## 问题2：输入cd,cat 这些常用命令加上空格就会断开

Image

## 当前版本

Image

binaricat:
## Root cause

Terminal autocomplete treated blank path arguments such as `cd ` and `cat ` as path-completion requests. On SSH sessions, that called `netcatty:ssh:listdir`, which opened an extra SSH `exec` channel and ran a remote `find` command on the existing connection. Strict bastion/audit gateways and network CLIs can close the connection when that hidden exec channel appears, so typing a space after path commands looked like the terminal input itself disconnected the session.

## Fix

Remote path completion no longer queries remote directories when the current path token is blank. Local terminals keep the previous behavior, and non-empty relative remote path prefixes such as `cat wo` still complete as before.

## Verification

- `node --test --import tsx components/terminal/completionEngine.test.ts`
- `npm run lint`
- `npm run build`

Fixes #1193

### DIFF
--- a/components/terminal/autocomplete/completionEngine.ts
+++ b/components/terminal/autocomplete/completionEngine.ts
@@ -29,6 +29,7 @@ import {
   shouldDoPathCompletion,
   getPathSuggestions,
   resolvePathComponents,
+  hasRemotePathLookupToken,
 } from "./remotePathCompleter";
 import { getSnippetSuggestions } from "./snippetCompleter";
 import type { Snippet } from "../../../domain/models";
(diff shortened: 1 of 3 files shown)


binaricat:
Closing this PR for now because the current issue report does not include enough bastion/audit gateway details to safely validate the fix. We should collect more information about the affected bastion behavior before changing Netcatty behavior here.

PR #1796, opened 2026-06-29
[codex] Fix host keyword highlight enablement

### LINKED ISSUE #1502: [Bug] 新增加的关键字高亮不生效了
### Operating system

Windows

### Netcatty version

1.1.40

### Affected area

SSH connection / terminal

### Can you reproduce it?

Always (100%)

### Steps to reproduce

1、在全局设置或者当前连接的关键字高亮中增加关键字
2、手动输入或者日志中出现新增加的关键字。不生效
3、重启应用或重新打开连接也不生效

### Expected behavior

新加的关键字正常高亮

### Actual behavior

不生效，只有以前用很早的版本增加的那些关键字正常使用，新加的不生效

### Logs / screenshots

_No response_

### Before submitting

binaricat:
## Root cause

In `components/terminal/HostKeywordHighlightPopover.tsx`, `handleAddRule` appended the new host-level rule but only turned on `keywordHighlightEnabled` when the host previously had zero rules. If a host already had saved rules while the host highlight switch was off, a newly added rule was saved and shown in the panel but never included by the terminal's rule merge in `useTerminalEffects`, because host rules are passed to the highlighter only when `host.keywordHighlightEnabled` is true. That explains the visible saved rule, the open rule row, and no terminal highlight for newly added current-connection rules.

The global path was checked separately: current `KeywordHighlighter.setRules` does highlight a newly added rule against visible terminal text, and this PR adds a regression test for that behavior too.

## Fix

- Add a single helper for adding host keyword highlight rules that always enables host-level highlighting.
- Use that helper from the current-connection highlight popover so the rule list and switch cannot diverge.
- Add regression tests for current-connection enablement and renderer-level matching of a newly added rule.

## Verification

- `node --test --import tsx components/terminal/HostKeywordHighlightPopover.test.ts components/terminal/keywordHighlight.test.ts components/terminal/keywordHighlightRegex.test.ts domain/keywordHighlight.test.ts`
- `npx eslint components/terminal/HostKeywordHighlightPopover.tsx components/terminal/HostKeywordHighlightPopover.test.ts components/terminal/keywordHighlight.test.ts components/terminal/keywordHighlight.ts components/terminal/keywordHighlightRegex.ts`
- `npm run build`
- `npm test` was attempted; it fails on unrelated existing/environment issues, including Electron binary installation errors and unrelated UI/memo tests, while all keyword-highlight targeted tests pass.

Fixes #1502

### DIFF
--- a/components/terminal/HostKeywordHighlightPopover.test.ts
+++ b/components/terminal/HostKeywordHighlightPopover.test.ts
@@ -1,0 +1,42 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+
+import type { Host, KeywordHighlightRule } from "../../types.ts";
+import { addHostKeywordHighlightRule } from "./HostKeywordHighlightPopover.tsx";
+
+const baseHost: Host = {
+  id: "host-1",
+  label: "Production",
+  hostname: "prod.example.com",
+  username: "root",
+  tags: [],
+  os: "linux",
+  keywordHighlightEnabled: false,
+  keywordHighlightRules: [
+    {
+      id: "old-rule",
+      label: "Old rule",
+      patterns: ["OLD"],
+      color: "#FBBF24",
+      enabled: true,
+    },
+  ],
+};
+
+const newRule: KeywordHighlightRule = {
+  id: "new-rule",
+  label: "Deploy",
+  patterns: ["DEPLOY"],
+  color: "#F87171",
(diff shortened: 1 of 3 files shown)


binaricat:
本地已跑 `codex review --base origin/main`，没有需要处理的问题。

PR #1804, opened 2026-06-29
Fix terminal history edit echo desync

### LINKED ISSUE #1768: [Bug] 使用方向键上，调出上次命令，只需编辑下命令（删减增加），命令就会莫名增加字符
### Operating system

Windows

### Netcatty version

1.1.47

### Affected area

SSH connection / terminal

### Can you reproduce it?

Always (100%)

### Steps to reproduce

使用方向键上，调出上次命令，只需编辑下命令（删减增加），命令就会莫名增加字符

### Expected behavior

使用方向键上，调出上次命令，修复自己打错的命令，正常提交

### Actual behavior

命令就会莫名增加字符

### Logs / screenshots

_No response_

### Before submitting

binaricat:
## Root cause
`components/terminal/runtime/terminalOutputPipeline.ts` `prioritizeTerminalInput()` treated deferred IPC ACK bytes as a reason to run the full stale-output drain. `createXTermRuntime.ts` calls that function before ordinary `term.onData` input, including Up-arrow history recall edits and Backspace. When a small readline/history echo such as `systemctl dddd` was still coalesced for the next render frame, any older deferred ACK made the function abort the pending display write and reset flow state. The remote shell still had the full line, but xterm missed part of the echo, so deleting/editing made the prefix appear undeletable and later echoes looked like duplicated or added characters.

## Fix
- When there is no visible output backlog and only deferred ACK bytes exist, flush the deferred ACK after the input is forwarded without aborting the pending display coalescer/write queue.
- Make interrupt stale-output draining opt-in via `drainStaleOutput === true`, preserving ordinary/direct-call display output unless the caller explicitly requested interrupt cleanup.
- Added a regression test that simulates a deferred ACK plus a pending `systemctl dddd` line-edit echo before ordinary input.

Fixes #1768

## Verification
- `node --test --import tsx components/terminal/runtime/terminalOutputPipeline.test.ts components/terminal/runtime/terminalSessionAttachment.test.ts components/terminal/runtime/terminalWriteAckDeferral.test.ts components/terminal/runtime/terminalWriteQueue.test.ts components/terminal/runtime/createXTermRuntime.test.ts`
- `node --test electron/bridges/terminalBridge.ctrlCInterrupt.test.cjs electron/bridges/terminalFlowAck.test.cjs`
- `npm run lint`
- `git diff --check`

I also ran full `npm test`. After running `npm run postinstall` to apply dependency patches, the remaining failures were outside this change area (theme/app-resume/snippet completer/terminal memo/capability codegen tests), while the terminal input/output and Ctrl+C flow tests above pass.

### DIFF
--- a/components/terminal/runtime/terminalOutputPipeline.test.ts
+++ b/components/terminal/runtime/terminalOutputPipeline.test.ts
@@ -11,6 +11,12 @@ import {
   teardownTerminalOutputPipeline,
 } from "./terminalOutputPipeline.ts";
 import { FLOW_LOW_WATER_MARK } from "./terminalFlowConstants.ts";
+import {
+  enqueueCoalescedTerminalWrite,
+  flushTerminalWriteCoalescer,
+  getTerminalWriteCoalescerPendingBytes,
+  resetTerminalWriteCoalescer,
+} from "./terminalWriteCoalescer.ts";
 import { enqueueTerminalWrite } from "./terminalWriteQueue.ts";
 import { accumulateDeferredTerminalWriteAck } from "./terminalWriteAckDeferral.ts";
 import { clearTerminalSessionFlowAck } from "./terminalFlowAckBuffer.ts";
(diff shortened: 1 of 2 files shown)


binaricat:
本地已运行 `codex review --base origin/main`，未发现需要处理的问题，复审已清空。

PR #1809, opened 2026-06-29
fix: Windows UI font fallback order

pyroch:
## Summary
- Keep only the bundled flag emoji subset before the selected UI font on Windows
- Move broad Segoe emoji/symbol fallbacks after the selected UI font
- Align the initial CSS font stack with the runtime UI font stack

## Why
`Segoe UI Emoji` / `Segoe UI Symbol` were placed before the selected UI font on Windows. One of those fonts can cover Latin glyphs, so English UI text ignored the selected interface font and rendered heavier/blurrier, while Cyrillic text continued to fall through to the selected/fallback text font.

`Noto Color Emoji Flags` is safe to keep first because it is unicode-range limited to regional indicator glyphs.

Before:
Netcatty_Zk1SGENBVy

After:
Netcatty_32us2loQZo

## Testing
- node --test --import tsx infrastructure/config/uiFonts.test.ts
- npm run lint -- infrastructure/config/uiFonts.ts index.css

### DIFF
--- a/index.css
+++ b/index.css
@@ -140,10 +140,10 @@
 
 }
 
-/* Windows: bundled flag font + Segoe UI Emoji ahead of Mona Sans so host
+/* Windows: bundled flag font ahead of Mona Sans so host
    labels render flag icons instead of ISO country codes (#1589, #1604). */
 html.platform-win32 {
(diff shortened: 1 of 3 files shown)


pyroch:
@codex review

PR #1814, opened 2026-06-29
fix: windows fullscreen control buttons hit area

pyroch:
**Summary**
Fix Windows frameless window controls so the top-right corner reliably hits the close button in fullscreen/maximized mode.

**Why**
The always-on top drag stripe covered the very top edge of the titlebar, including the window-controls area. On Windows, that made the extreme top-right corner resolve as a drag region instead of the close button, so hover/click did not work there. The fix keeps the drag stripe out of the controls area and makes the controls fill the titlebar height while preserving their existing width.

**Validation**
node --test --import tsx components\TopTabs.test.ts
git diff --check

Before:
Netcatty_veoUm6ea9p
After:
Netcatty_bxiqISeLBJ


pyroch:
@binaricat hey, could you check this? In my opinion that's pretty good QoL UX for Windows users.

binaricat:
Hey @pyroch, really appreciate you digging into this — the hit area issue on Windows is genuinely annoying and it's great to see someone tackling it! 🙌

One thing worth knowing: the layout in your "After" screenshot is actually what we had in an earlier version of the app. We intentionally moved away from it at the time because the control buttons ended up not being visually bottom-aligned with the toolbar row, which felt a bit off from a UI standpoint.

So the tricky part here is that the fix solves the hit area problem but brings back that old visual inconsistency. I'd love to find a solution that gets the best of both worlds!

Once I get through some of the bigger performance issues I'm working on right now, I'll definitely come back and give this a proper think — it deserves a clean solution. Thanks again for the contribution, it's really appreciated! 🙏

Issue #1815, opened 2026-06-29
[Bug] 开启会话日志后未记录内容


Meow-Emperor:
### Operating system

macOS

### Netcatty version

v1.1.47

### How did you install Netcatty?

GitHub Release (.dmg / .exe / .AppImage / .deb / .rpm / .pacman)

### Affected area

Other

### Can you reproduce it?

Always (100%)

### Steps to reproduce

连接上服务器后点击右上角的开始会话日志，然后开始输命令 结束会话日志后日志文件什么都没有记录

### Expected behavior

应该记录输入的命令和命令的对应输出

### Actual behavior

什么也没记录

### Before submitting

- [x] I searched existing issues and did not find a duplicate
- [x] I removed passwords, private keys, and other secrets from this report

Meow-Emperor:
并没有使用ctrl+c终断而是等待输出完成后才关闭，但是内容还是只有机器名称

PR #1816, opened 2026-06-29
[codex] Fix session log capture during interrupt drain

binaricat:
## What changed
- Writes interrupt-drained pending terminal output to the active session log stream.
- Extends the Ctrl+C interrupt test to assert that the drained output is logged as well as rendered.

## Why
Issue #1815 reports that manual session logs can end up empty after starting a log on a connected session and running commands. I found a concrete gap introduced around the interrupt-drain path: when pending output was flushed to the terminal during high output pressure, it was not also appended to the session log stream.

## Validation
- `node --test electron/bridges/terminalBridge.ctrlCInterrupt.test.cjs`
- `node --test electron/bridges/sessionLogStreamManager.test.cjs electron/bridges/sessionLogsBridge.test.cjs electron/bridges/emitTerminalSessionData.test.cjs`
- `npx eslint electron/bridges/terminalBridge.cjs electron/bridges/terminalBridge.ctrlCInterrupt.test.cjs`

### DIFF
--- a/electron/bridges/terminalBridge.cjs
+++ b/electron/bridges/terminalBridge.cjs
@@ -1015,7 +1015,6 @@ function drainPendingOutputForInterrupt(sessionId, session, trace) {
     cols: session.cols,
     rows: session.rows,
   });
-  sessionLogStreamManager.appendData(sessionId, output.data);
 }
 
 function interruptSession(event, payload) {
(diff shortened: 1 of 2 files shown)


binaricat:
Closing this PR after review: the proposed append would duplicate already-logged interrupt-drained output, so there is no safe fix here without a clearer root cause for #1815.

PR #1831, opened 2026-06-29
Fix Telnet identity credential resolution

binaricat:
## Summary
- Stop Telnet from falling back to SSH username/password when Telnet credentials are unset
- Let Telnet use the selected Keychain identity username/password
- Update Telnet startup and credential-copy tests

## Verification
- node --test --import tsx components/terminal/runtime/createTerminalSessionStarters.test.ts components/terminal/runtime/createTerminalSessionStarters.telnet.test.ts domain/host.test.ts application/AppHandlers.connect.test.ts
- npx eslint domain/host.ts domain/host.test.ts components/terminal/runtime/createTerminalSessionStarters.ts components/terminal/runtime/createTerminalSessionStarters.telnet.test.ts components/VaultView.tsx
- npm run build

### DIFF
--- a/components/VaultView.tsx
+++ b/components/VaultView.tsx
@@ -631,11 +631,11 @@ const VaultViewInner: React.FC<VaultViewProps> = ({
       : undefined;
 
     const username = isTelnet
-      ? resolveTelnetUsername(effective)
+      ? resolveTelnetUsername(effective, identity)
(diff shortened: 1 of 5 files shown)


binaricat:
Closing this PR because the requested change was sent to the wrong place.

PR #1837, opened 2026-06-30
fix(snippets): preserve multi-line shell commands when auto-running

### LINKED ISSUE #1836: [Bug] 多行脚本运行可能会丢失命令
### Operating system

Linux

### Netcatty version

1.1.51

### Affected area

Snippets

### Can you reproduce it?

Always (100%)

### Steps to reproduce

①创建代码片段
```
sudo apt install gconf2-common -y
echo "123456"
```
②点击运行

### Expected behavior

命令按顺序执行，最后会打印123456。就像直接在终端一次性粘贴这两行，最后应打印123456。
 
手动粘贴两行运行:
Image

### Actual behavior

没有打印123456

点击脚本运行:
Image

### Logs / screenshots

如上

### Before submitting

binaricat:
## Summary

Fixes #1836

Multi-line snippet auto-run used a fixed inter-line delay for every connection type. When the first command was still running (e.g. `sudo apt install ...`), later lines were injected into the foreground process instead of being queued by the shell, so trailing commands like `echo "123456"` were lost.

- **Shell sessions (SSH/local/etc.)**: send multi-line startup snippets in one write so the shell can queue commands, matching manual paste behavior.
- **Connected terminals**: only use inter-line delays for telnet prompt sequences; SSH sessions use bracketed-paste batching again.
- **Telnet**: keep the existing line-by-line delay for login prompt sequences.

## Test plan

- [x] `node --test --import tsx components/terminalHelpers.test.ts`
- [x] `node --test --import tsx components/terminal/runtime/createTerminalSessionStarters.test.ts` (new regression tests for shell batch send + telnet sequencing)
- [ ] Manually run a two-line snippet (`sudo apt install ...` + `echo "123456"`) and confirm the second line executes after the first finishes

### DIFF
--- a/components/Terminal.tsx
+++ b/components/Terminal.tsx
@@ -1752,7 +1752,7 @@ const TerminalComponent: React.FC<TerminalProps> = ({
     if (!term || !id) return;
 
     let data = normalizeLineEndings(command);
-    const lineDelayMs = shouldDelayAutoRunSnippetInput(data, { noAutoRun })
+    const lineDelayMs = shouldDelayAutoRunSnippetInput(data, { noAutoRun, protocol: host.protocol })
       ? AUTO_RUN_SNIPPET_LINE_DELAY_MS
       : undefined;
     const isMultiLine = data.includes('\n');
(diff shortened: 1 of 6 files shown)


binaricat:
Closing per maintainer request — will revisit with a settings-based approach instead of changing default behavior.

Issue #1850, opened 2026-06-30
[Bug] Agent command execution can terminate the active SSH session when shell exits


RoryChou-flux:
### Operating system

macOS

### Netcatty version

1.1.51

### How did you install Netcatty?

GitHub Release (.dmg / .exe / .AppImage / .deb / .rpm / .pacman)

### Affected area

SSH connection / terminal, AI assistant

### Can you reproduce it?

Always (100%)

### Steps to reproduce

1. Open an SSH session in Netcatty.
2. Run an Agent command that executes a multi-line shell script in the active SSH session.
3. Example command:

```bash
set -e
cd /opt/actions-runner/seoptimize
./config.sh --version
test -x ./config.sh && echo 'config.sh ok'
test -x ./run.sh && echo 'run.sh ok'
test -d ./bin && echo 'bin ok'
test -d ./externals && echo 'externals ok'
stat -c '%U:%G %n' . config.sh run.sh

### Expected behavior

The Agent should not cause the active SSH terminal session to be terminated just because a generated command exits with a non-zero status.

Ideally, Netcatty should either:

- execute agent-generated shell commands inside an isolated child shell instead of the user’s active login shell;
- wrap generated multi-line commands with a failure guard, for example:

```bash
bash -lc '
set -e
cd /opt/actions-runner/seoptimize
./config.sh --version
test -x ./config.sh && echo "config.sh ok"
test -x ./run.sh && echo "run.sh ok"
test -d ./bin && echo "bin ok"
test -d ./externals && echo "externals ok"
stat -c "%U:%G %n" . config.sh run.sh
' || echo "Agent command failed; SSH session preserved."
```

- warn before running commands that may terminate the active shell, such as commands containing `set -e`, `exit`, `logout`, `exec`, `kill $$`, or similar;
- provide a safer prefilled command pattern for diagnostics, especially when the Agent generates multi-line shell snippets.

### Actual behavior

The SSH session closes immediately or appears to disconnect when the Agent runs the command and one of the unguarded commands returns a non-zero exit code.

In this case, the command included `set -e`, so when a command such as `cd`, `./config.sh --version`, or `stat` fails, the shell exits. Because the Agent appears to execute the generated command in the active SSH shell, the whole SSH session/window is terminated instead of only the Agent command failing.

Example command that triggered the issue:

```bash
set -e
cd /opt/actions-runner/seoptimize
./config.sh --version
test -x ./config.sh && echo 'config.sh ok'
[...]

PR #1852, opened 2026-07-01
[codex] Isolate agent shell commands

binaricat:
## Summary
- Prevent AI shell commands from killing or exiting the active terminal before Netcatty can emit the completion marker.
- Preserve ordinary state changes such as cd/export/source when they are safe.
- Block or isolate risky shell paths including sourced files, helper redefinitions, shell options, traps, aliases/functions, redirections, path-prefixed kill commands, and signal-option kill variants.

## Verification
- node --test --test-name-pattern "dynamic shell-terminating|dangerous existing aliases|indirect self-kill|assignment-prefixed path kill" electron/bridges/ai/ptyExec.test.cjs
- node --test electron/bridges/ai/ptyExec.test.cjs
- node --test electron/terminalWorker/aiExec.test.cjs electron/bridges/aiBridge/cattyExecHandlers.worker.test.cjs electron/bridges/aiBridge.test.cjs
- npm run lint
- git diff --check
- npm test (3983 pass, 0 fail, 3 skip)

## Review note
- Independent review found source/function/alias kill bypasses; this branch now includes regression tests for assignment-prefixed path kill plus `-s`, `--signal`, `--signal=`, `-n`, and variable-target kill forms, and the full local suite passes.

### DIFF
--- a/electron/bridges/ai/ptyExec.test.cjs
+++ b/electron/bridges/ai/ptyExec.test.cjs
@@ -1,7 +1,7 @@
 const test = require("node:test");
 const assert = require("node:assert/strict");
 const { spawnSync } = require("node:child_process");
-const { mkdtempSync, rmSync } = require("node:fs");
+const { mkdtempSync, rmSync, symlinkSync, writeFileSync } = require("node:fs");
 const { tmpdir } = require("node:os");
 const { join } = require("node:path");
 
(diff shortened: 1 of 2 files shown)


binaricat:
Closing this PR because the shell-isolation patch grew too broad and regressed AI behavior. We'll revisit issue #1850 with a smaller, safer approach.

PR #1872, opened 2026-07-01
[codex] Fix terminal tab switch recovery

binaricat:
## Summary

Fix terminal tab switching so background terminal panes keep their own sizing context and do not briefly inherit the active tab's AI-panel width. The recovery path now restores newly visible terminals immediately, while hidden terminals skip background refits from output, connection completion, and font loading.

## Why

A hidden tab could be affected by another tab's side panel width, especially while heavy terminal output was streaming. When switching back, the terminal could paint with stale geometry before the follow-up fit corrected it, leaving a visible blank area for a moment.

## Validation

- `node --test --import tsx components/terminal/terminalTabVisibilityRecovery.test.ts components/terminal/terminalSafeFitVisibility.test.ts components/terminal/runtime/terminalPostConnectVisibility.test.ts components/terminalPaneVisibility.test.ts`
- `npm run lint`
- `npm test`
- `npm run build`
- `git diff --check`

### DIFF
--- a/components/Terminal.tsx
+++ b/components/Terminal.tsx
@@ -1623,10 +1623,13 @@ const TerminalComponent: React.FC<TerminalProps> = ({
     });
   }, [reuseConnectionFromSessionId, sessionId, terminalBackend]);
 
-  const safeFit = (options?: { force?: boolean; requireVisible?: boolean }) => {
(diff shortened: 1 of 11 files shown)


binaricat:
@codex review

PR #1882, opened 2026-07-02
[codex] Keep agent set -e failures from closing shells

### LINKED ISSUE #1850: [Bug] Agent command execution can terminate the active SSH session when shell exits

binaricat:
## Summary

- Apply the smallest shell-wrapper change for issue #1850.
- When an Agent command enables `set -e` / `set -o errexit`, run that command in a subshell so a command failure does not close the active SSH/login shell.
- If the active shell already has errexit enabled, temporarily relax it around the wrapper and restore it afterward.
- Keep normal stateful commands, such as `cd`, running in the active shell as before.

## Intentional scope

This PR is deliberately not a broad shell-isolation rewrite. The previous attempt grew into detecting many shell-terminating patterns and regressed normal AI behavior. This version only targets the simplest high-value path behind #1850: `set -e` failures taking down the active session.

It does not try to solve every possible way a command could exit or kill a shell.

Fixes #1850

## Validation

- `node --test electron/bridges/ai/ptyExec.test.cjs electron/terminalWorker/aiExec.test.cjs electron/bridges/aiBridge/cattyExecHandlers.worker.test.cjs`
- `npx eslint electron/bridges/ai/ptyExecHelpers.cjs electron/bridges/ai/ptyExec.test.cjs`
- `git diff --check`

### DIFF
--- a/electron/bridges/ai/ptyExec.test.cjs
+++ b/electron/bridges/ai/ptyExec.test.cjs
@@ -1,5 +1,9 @@
 const test = require("node:test");
 const assert = require("node:assert/strict");
+const { spawnSync } = require("node:child_process");
+const { mkdtempSync, rmSync } = require("node:fs");
+const { tmpdir } = require("node:os");
+const { join } = require("node:path");
 
 const {
   resolveEffectiveShellKind,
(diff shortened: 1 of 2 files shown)


binaricat:
Closing this draft because the minimal set-e detection approach started expanding into shell parsing. We'll revisit with a simpler scoped approach.

Issue #1914, opened 2026-07-03
[Bug] Two packaging bugs in netcatty-1.1.53-1.x86_64


danielabelski:
### Operating system

Linux

### Netcatty version

1.1.53

### How did you install Netcatty?

GitHub Release (.dmg / .exe / .AppImage / .deb / .rpm / .pacman)

### Affected area

Other

### Can you reproduce it?

Always (100%)

### Steps to reproduce

### 1. RPM claims ownership of `.build-id` symlinks it shouldn't own   
Installing Slack 4.50.143 fails with: 

file /usr/lib/.build-id/42/f633ec... from install of slack-4.50.143-0.1.el8.x86_64
conflicts with file from package netcatty-1.1.53-1.x86_64
(+ 3 more .build-id conflicts)                                                                                                                                                                                                                                                                                                                              

Those SHA1 hashes come from Slack's ELF binaries, so netcatty can't legitimately own them. This usually             means the spec has something like `/usr/lib/.build-id/*` or `/usr/lib/.build-id` in `%files`.                                                                                                                                                       

Workaround: `sudo rpm -Uvh --replacefiles <other-package>.rpm`.                                                                                                                                                               

### 2. Shared libraries shipped as regular files instead of symlinks                                                                                                                                                                    Every RPM transaction on a system with netcatty installed prints:                                                                                                                                                                                                                                                                                           ldconfig: /lib/libtranscribe.so.0 is not a symbolic link
ldconfig: /lib/libggml.so.0 is not a symbolic link
ldconfig: /lib/libggml-base.so.0 is not a symbolic link                                                                                                                                                                                                                                                                                                     

The `SONAME` file (e.g. `libggml.so.0`) must be a symlink to the fully versioned file (`libggml.so.0.X.Y`), not a regular file. Ship both in `%files`:                                                                                                                                                                                                                                                                                                                  %{_libdir}/libggml.so.0.X.Y
%{_libdir}/libggml.so.0        # symlink                                                                                                                                                                                                                                                                                                                    

Same for `libtranscribe` and `libggml-base`. Also, on Fedora/RHEL x86_64 these should live in `/usr/lib64`,  not `/lib`.

### 3.  Environment
- Fedora, x86_64
- netcatty 1.1.53-1                  

### Expected behavior

Installation with no errors

### Actual behavior

subj

### Logs / screenshots

_No response_

### Before submitting

- [x] I searched existing issues and did not find a duplicate
[...]

PR #1919, opened 2026-07-04
Fix RPM build-id file conflicts

binaricat:
## Summary
- Disable RPM build-id symlink generation so Netcatty RPMs do not own `/usr/lib/.build-id` entries that can conflict with other RPM packages.
- Add an RPM artifact verifier that checks the package is readable, has the expected architecture, and does not include `/usr/lib/.build-id`, `libggml`, or `libtranscribe` entries.
- Run the RPM verifier in both Linux package CI jobs.

## Validation
- `npm ci`
- `node --test scripts/electron-builder-config.test.cjs scripts/github-workflow-build.test.cjs`
- `npm run build`
- Verified the new RPM script fails against the released `Netcatty-1.1.53-linux-x86_64.rpm` because it contains `/usr/lib/.build-id` entries.
- Built an ARM64 RPM in a Linux Docker container and verified it passes `scripts/verify-linux-rpm-artifact.sh aarch64`.

## Notes
Refs #1914. This fixes the confirmed `.build-id` ownership conflict. I did not change anything for the `libggml` / `libtranscribe` ldconfig messages because those files were not present in the checked Netcatty RPM file list; the new verifier also guards against accidentally shipping them.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -411,8 +411,8 @@ jobs:
       - name: Install deps
         run: npm ci
 
-      - name: Install pacman packaging dependencies
-        run: sudo apt-get update && sudo apt-get install -y libarchive-tools
+      - name: Install Linux packaging dependencies
+        run: sudo apt-get update && sudo apt-get install -y libarchive-tools rpm
 
       - name: Set version
         shell: bash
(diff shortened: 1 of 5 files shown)


binaricat:
@codex review

PR #1924, opened 2026-07-04
[codex] Add enter reconnect for disconnected terminals

binaricat:
## Summary
- allow a disconnected terminal to reconnect when the user presses plain Enter
- return focus to the terminal after the disconnected notice is dismissed so Enter still works
- avoid stealing Enter from search, compose, auth, host-key prompts, transfers, overlays, or controls outside xterm
- add coverage for reconnect key gating, interactive-target gating, and dismiss-focus behavior

## Context
Addresses the reconnect pain discussed in #1718, especially the request for Enter to reconnect after a session is lost.

## Validation
- node --test --import tsx components/terminal/TerminalView.test.tsx components/terminal/TerminalConnectionDialog.test.ts components/terminal/TerminalContextMenu.test.ts components/terminalHelpers.test.ts components/terminal/restoredSessionGate.test.ts application/state/resolveTerminalSessionExitIntent.test.ts
- npx eslint components/Terminal.tsx components/terminal/TerminalView.tsx components/terminal/TerminalView.test.tsx components/terminal/restoredSessionGate.test.ts
- git diff --check
- npm run build


binaricat:
@codex review

PR #1933, opened 2026-07-04
[codex] Fix manual session log capture

### LINKED ISSUE #1815: [Bug] 开启会话日志后未记录内容

binaricat:
## What changed
- Save manual toolbar session logs as raw `.log` streams instead of rendered text snapshots.
- Keep the existing prompt attached to ordinary command echoes while separating device output that starts with a carriage-return redraw.
- Avoid adding extra blank lines when the next terminal data already starts with a normal newline, including when that newline arrives split across chunks.
- Flush a pending first carriage return on stop so no terminal data is lost.
- Keep manual session-log saves on `.log` files, even if the selected path has another extension.
- Ask before overwriting an existing normalized `.log` path when the save dialog confirmed a different selected path.
- Add regression coverage for network-device carriage-return output, ordinary shell command echo output, leading newline output, split leading newline output, stop-after-carriage-return output, and normalized-path overwrite confirmation.

## Root cause
Manual session logging was added as a readable text snapshot path even though the save dialog defaults to `.log` and the toolbar action is expected to capture the full session stream. Terminal carriage-return rewrites and prompt redraws can collapse the rendered snapshot, leaving the saved file with only the prompt/hostname instead of the commands and output.

Fixes #1815.

## Validation
- `node --test electron/bridges/sessionLogsBridge.test.cjs electron/bridges/sessionLogStreamManager.test.cjs`
- `npx eslint electron/bridges/sessionLogsBridge.cjs electron/bridges/sessionLogsBridge.test.cjs electron/bridges/sessionLogStreamManager.cjs electron/bridges/sessionLogStreamManager.test.cjs`
- `git diff --check`
- `npm run build`

### DIFF
--- a/electron/bridges/sessionLogStreamManager.cjs
+++ b/electron/bridges/sessionLogStreamManager.cjs
@@ -232,6 +232,8 @@ function createStreamEntry(sessionId, opts) {
     disabled: false,
     startToken,
     stopRequiresToken: Boolean(opts.stopRequiresToken),
+    separateInitialLineBeforeLeadingCarriageReturn: false,
+    pendingInitialLineLeadingCarriageReturn: false,
   };
 
   entry.flushTimer = setInterval(() => {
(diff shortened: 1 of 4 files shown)


binaricat:
@codex review

Issue #1761, opened 2026-06-28
[Bug] 终端窗口失去焦点后tmux刷新断开


Kaoruha:
### Operating system

Windows

### Netcatty version

1.1.47

### How did you install Netcatty?

GitHub Release (.dmg / .exe / .AppImage / .deb / .rpm / .pacman)

### Affected area

SSH connection / terminal

### Can you reproduce it?

Always (100%)

### Steps to reproduce

1. 运行 NetCatty 连接一个主机
2. 运行 tmux
3. 鼠标点击NetCatty外部

### Expected behavior

会话保持，时刻拿到返回

### Actual behavior

NetCatty窗口失去焦点后，会话无法刷新，重新聚焦后需要一段时间才能有响应，想问一下这是bug还是节能设计

### Before submitting

- [x] I searched existing issues and did not find a duplicate
- [x] I removed passwords, private keys, and other secrets from this report

PR #1940, opened 2026-07-05
Fix terminal catch-up lag when window is backgrounded unfocused (#1880)

### LINKED ISSUE #1880: [Bug] 后台切换至前台日志卡顿
### Operating system

Windows

### Netcatty version

1.1.54

### Affected area

SSH connection / terminal

### Can you reproduce it?

Always (100%)

### Steps to reproduce

1. 连接至tmux窗口
2. 监控长日志输出
3. 将netcatty窗口放置后台较长时间
4. 切换至前台

### Expected behavior

应该立即刷新到最新的日志

### Actual behavior

需要经历缓慢的，每秒刷新间隔约1分钟的日志记录后，直至最新
例如：12：00切换至后台
13：00打开，则会经历12：01，12：02……一直刷新

### Logs / screenshots

_No response_

### Before submitting

binaricat:
## Summary

- **Root cause (#1880)**: PR #1881 only bypassed rAF/write-queue throttling when `document.visibilityState === 'hidden'`. On Windows, Alt+Tabbing away often leaves the page `visible` but unfocused; Chromium throttles `requestAnimationFrame`, so the write coalescer backlog grows. Returning to the app then replays tmux/TUI frames slowly (12:01, 12:02, …) and blocks the UI thread (black screen + button lag).
- **Fix**: Treat unfocused-but-visible windows like hidden pages for synchronous terminal writes; skip debounced repaints during that fast path; flush any remaining coalescer/queue/xterm buffer on app resume (`visibilitychange`, `focus`, `onWindowShown`).
- **Related (#949 local nvtop overnight freeze)**: Same class of issue for long-running TUIs — background repaints + backlog were burning CPU/memory. This should reduce overnight pressure; full freeze may still need separate investigation if it persists.

## Test plan

- [x] `node --test --import tsx components/terminal/runtime/terminalUnfocusedRepaint.test.ts`
- [x] `node --test --import tsx components/terminal/runtime/terminalSessionAttachment.test.ts`
- [x] `node --test --import tsx components/terminal/appResumeWebglRecovery.test.ts`
- [ ] Manual: connect to tmux tailing logs, Alt+Tab away 30+ min, return — should show latest output immediately without minute-by-minute replay
- [ ] Manual: run nvtop/htop locally, leave app unfocused overnight — verify app stays responsive on return

Fixes #1880

### DIFF
--- a/application/state/useTerminalBackend.ts
+++ b/application/state/useTerminalBackend.ts
@@ -168,6 +168,11 @@ export const useTerminalBackend = () => {
     return bridge?.onWindowFullScreenChanged?.(cb);
   }, []);
 
+  const onWindowShown = useCallback((cb: () => void) => {
+    const bridge = netcattyBridge.get();
+    return bridge?.onWindowShown?.(cb);
+  }, []);
+
   const onHostKeyVerification = useCallback((cb: Parameters<NonNullable<NetcattyBridge["onHostKeyVerification"]>>[0]) => {
     const bridge = netcattyBridge.get();
     return bridge?.onHostKeyVerification?.(cb);
@@ -370,6 +375,7 @@ export const useTerminalBackend = () => {
       onChainProgress,
       onConnectionReuseFallback,
       onWindowFullScreenChanged,
+      onWindowShown,
       onHostKeyVerification,
       respondHostKeyVerification,
       openExternal,
@@ -422,6 +428,7 @@ export const useTerminalBackend = () => {
       onChainProgress,
       onConnectionReuseFallback,
       onWindowFullScreenChanged,
+      onWindowShown,
       onHostKeyVerification,
       respondHostKeyVerification,
       openExternal,
--- a/components/terminal/appResumeWebglRecovery.test.ts
+++ b/components/terminal/appResumeWebglRecovery.test.ts
@@ -2,15 +2,12 @@ import test from "node:test";
 import assert from "node:assert/strict";
 import { readFileSync } from "node:fs";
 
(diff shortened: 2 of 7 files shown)


binaricat:
Second-pass root-cause verification (per review request):

**Confirmed mechanism** — #1881 only bypasses rAF/timer throttling when `document.visibilityState === 'hidden'`. On Windows, Chromium's native occlusion tracking (`CalculateNativeWinOcclusion`) backgrounds the renderer of an occluded window (rAF stops, timers throttled to 1Hz / 1-min intensive throttling) while Electron may still report the page as `visible` (electron/electron#29282). Repo history independently confirms this state exists: commit 232cbb9b (fix for #1761) explicitly targets "visible but unfocused" write-callback stalls on Windows. In that state 1.1.56 still routes output through the throttled 67ms flush timer → backlog grows → slow replay on foreground return, with no flush on resume. That matches the 1.1.56 reproduction in #1880.

**Two defects found in the first commit of this PR, now fixed in 7ed40da0:**
1. `terminalBackend.onWindowShown` did not exist on the `useTerminalBackend` surface, so the window-shown resume flush silently never fired. Now exposed and covered by a source-level test.
2. The background fast path skipped `scheduleTerminalRepaintWhenUnfocused`, which would have re-frozen visible-but-unfocused windows (direct #1761 regression) since that debounced sync repaint is the only render path without rAF. Restored; it already no-ops when the page is truly hidden.

Full suite: 4216/4222 pass; the 3 failures are pre-existing `ssh2CertificateAgentPatch` failures reproducible on the base branch without these changes.

PR #1972, opened 2026-07-06
fix(ui): replace native confirm() with an in-app dialog

LaiYueJi:
## Problem
Destructive actions (delete host, delete identity, remove AI provider, Docker container/image remove, tmux kill, …) use the native `window.confirm()` / `globalThis.confirm()`. On Windows the native dialog can leave the Electron renderer with broken keyboard focus / modal state: after it is dismissed **the next input box silently ignores typing** until the tree re-renders (switching UI language or restarting "fixes" it). This is the same failure `SystemPanelConfirmDialog` was already created to avoid — but most call sites still use the native confirm.

Repro: right-click a host → **Delete** → confirm → the search / input box can no longer be typed into.

## Change
- Add a promise-based `useConfirm()` / `ConfirmProvider` (`components/ui/confirm.tsx`) that renders an in-app Radix dialog (same look as `SystemPanelConfirmDialog`).
- Wire `ConfirmProvider` into every window root that has its own `I18nProvider`: main app, settings, terminal-popup, tray.
- Migrate all native `confirm()` call sites: App host delete; KeychainManager identity delete (×2); settings (AI provider remove, quick-message delete, SFTP file-association remove); systemManager (Docker container remove/kill, image remove/prune, process signal/kill, tmux kill session/window).
- Opening the dialog is deferred to the next task so a triggering Radix **ContextMenu / DropdownMenu** finishes closing first — otherwise the two dismissable layers conflict and the dialog opens then instantly dismisses (overlay flashes, content never shows). This was the host-delete case, which is triggered from a context menu.
- `useConfirm()` falls back to native `confirm` when rendered without a provider, so isolated component tests keep passing unchanged.
- Adds `common.confirm` to en / ru / zh-CN / zh-TW.

## Verification
- `npm run build` passes; `eslint` clean on changed files; existing tests pass.
- Manually verified in a packaged Windows build: host delete now shows the in-app dialog and the input box keeps working afterward.

### DIFF
--- a/App.tsx
+++ b/App.tsx
@@ -57,6 +57,7 @@ import {
 import { getEffectiveKnownHosts } from './infrastructure/syncHelpers';
 import { ToastProvider, toast } from './components/ui/toast';
 import { TooltipProvider } from './components/ui/tooltip';
+import { ConfirmProvider, useConfirm } from './components/ui/confirm';
 import { PortForwardHostKeyDialog } from './components/port-forwarding';
 import { VaultSection } from './components/VaultView';
 import { KeyboardInteractiveRequest } from './components/KeyboardInteractiveModal';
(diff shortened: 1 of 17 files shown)


LaiYueJi:
Superseded by #1969 (already merged into main), which fixes the confirm-dialog input lock. Closing this in favor of that.

Issue #1992, opened 2026-07-07
[Bug] 进入日志文件  tail -n 100 -f xxx.log 日志文件10MB+  特别卡 直接卡死


luckamuu:
### Operating system

Windows

### Netcatty version

1.1.51

### How did you install Netcatty?

GitHub Release (.dmg / .exe / .AppImage / .deb / .rpm / .pacman)

### Affected area

SSH connection / terminal

### Can you reproduce it?

Always (100%)

### Steps to reproduce

进入日志文件  tail -n 100 -f xxx.log 日志文件10MB+

### Expected behavior

流畅

### Actual behavior

直接卡死动不了

### Logs / screenshots

Reported from Netcatty Settings (netcatty 1.1.51).

User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) netcatty/1.1.51 Chrome/148.0.7778.218 Electron/42.3.3 Safari/537.36

### Before submitting

- [x] I searched existing issues and did not find a duplicate
- [x] I removed passwords, private keys, and other secrets from this report

binaricat:
感谢反馈。这个问题报告来自 Netcatty 1.1.51，版本已经比较旧了；当前最新正式版是 v1.1.59，建议先升级到最新正式版。

我们今天已经合并了针对大文件 / `tail -f` 输出卡顿的优化：#1994 和 #1996。后续包含这些优化的版本发布后，也请再验证一下这个 `tail -n 100 -f` 10MB+ 日志场景；如果升级后仍然卡死，欢迎继续补充复现信息。

PR #1996, opened 2026-07-07
Follow up terminal tail performance fixes

binaricat:
Follow-up to #1994. This PR contains the review follow-up work that was pushed after #1994 had already been merged.

Summary:
- Keep hidden terminal panes from flushing large pending output until visible again.
- Flush pending hidden terminal writes before hibernating so buffered output is not lost.
- Preserve terminal performance metadata only when it still maps to a single original chunk.
- Make timestamp batching safer for wide characters, combining marks, and terminal control sequences.
- Preserve terminal output metadata through preload backlog paths when safe.

Verification:
- git diff --check origin/main...HEAD
- targeted terminal/preload tests: 136 passed
- rerun of previously failing Electron subset after local Electron install completed: 119 passed
- npm test: 4397 passed, 0 failed, 3 skipped
- npm run build

Refs #1992
Refs #1994

### DIFF
--- a/components/Terminal.tsx
+++ b/components/Terminal.tsx
@@ -155,6 +155,7 @@ import {
   flushTerminalSessionFlowAck,
 } from "./terminal/runtime/terminalFlowAckBuffer";
 import { releaseTerminalFlowBeforeHibernate } from "./terminal/runtime/terminalSessionAttachment";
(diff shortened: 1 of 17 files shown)


binaricat:
@codex review

PR #2045, opened 2026-07-09
feat: show live terminal hosts in SFTP picker

binaricat:
## Summary
- Add a Connected section to the SFTP host picker so users can jump straight to currently live SSH sessions without searching the full vault list (#2040)
- Reuse the selected terminal session via `sourceSessionId`, including ephemeral deep-link hosts
- Wire the same picker into both the top-level SFTP tab and the in-session side panel, while keeping vault writes on persisted hosts only

## Test plan
- [x] Connect 1–2 SSH hosts, open top-level SFTP → Select host, confirm Connected section appears
- [x] Pick a Connected host and confirm SFTP opens by reusing the live session (faster / no full re-auth when reuse works)
- [ ] Repeat from the terminal SFTP side panel host picker
- [ ] Open a password/deep-link ephemeral host, confirm it appears in Connected and can open SFTP
- [ ] Confirm serial/local/telnet sessions do not appear in Connected
- [ ] Confirm selecting a normal vault host still works, and ephemeral hosts are not written into vault


binaricat:
Addressed the Codex P2: `sessions` is now part of `SIDE_PANEL_STABLE_CTX_KEYS`, with a regression test so the Connected picker refreshes when terminal sessions change while the side panel is mounted.

binaricat:
@codex review

binaricat:
Addressed the latest Codex P2s:

1. Side-panel stable ctx still tracks `sessions`, but equality now uses `sftpPickerSessionsEqual` (id/hostId/protocol/status only), so OSC title/cwd churn no longer invalidates mounted side panels.
2. Applied the same comparison in `SftpView` / `SftpSidePanel` memo helpers.
3. Ignored local `.omo/` run-continuation state in `.gitignore`.

Local `codex review --base main` is clean (lint/tests/build passed).

binaricat:
@codex review

binaricat:
Addressed the latest Codex P2: when multiple live terminals share a hostId and status, `listSftpConnectedHosts` now keeps the later session so SFTP reuse prefers the newest connection (still prefers `connected` over a later `connecting`).

binaricat:
@codex review

binaricat:
Addressed the latest Codex P2: Connected picker now only lists reusable plain SSH sessions (`status === connected`, no Mosh/ET). Connecting / Mosh / ET sessions are excluded so selecting a row does not fall back to a fresh SFTP login / second MFA prompt.

binaricat:
@codex review

binaricat:
Addressed the latest Codex P2: Connected picker now decides reuse from **session** transport flags only. Vault hosts that default to Mosh/ET no longer hide a live plain-SSH terminal (e.g. `ssh://` deep links that clear those flags while keeping the saved hostId).

binaricat:
@codex review

binaricat:
Addressed the latest Codex P2s in `da1e9547` / `fc6a686a` / `48157f56`:

1. Connected rows overlay the live session endpoint (`hostname`/`username`/`port`) so `sourceSessionId` reuse matches `findReusableSession`.
2. Hosts with `sftpSudo` are excluded from Connected (bridge never reuses for sudo).
3. Saved hosts are only hidden when their endpoint matches the Connected row — edited vault hosts stay selectable alongside Live.

Local `codex review --base main` is clean after these fixes.

binaricat:
@codex review

binaricat:
Addressed the latest remote Codex P2s (and local review follow-ups) in `609c664e`…`78554e6d`:

1. Live Connected rows try `reuseOnly` openSftp with endpoint-only options **before** `getHostCredentials`, so missing/undecryptable vault secrets no longer block reuse.
2. `reuseOnly` fails closed (no incomplete fresh dial) and skips renderer endpoint matching so auth-dialog / identity usernames and default ports cannot reject a named live session.
3. Connected picker port defaults to live `22` when `session.port` is undefined (not edited vault port).
4. Memoization compares sessions positionally so same-host “latest session” preference stays correct after reorder.
5. zh-CN/zh-TW Connected section label: 已连接 / 已連線.

Local `codex review --base main` is clean after these.

PR #2081, opened 2026-07-10
Measure actual SSH network latency

### LINKED ISSUE #2074: [Bug] 终端显示的延迟不准确，偏高很多
### Operating system

Windows

### Netcatty version

1.1.60

### Affected area

SSH connection / terminal

### Can you reproduce it?

Always (100%)

### Steps to reproduce

内网的服务器（互联网的也是一样），通过ping和tcping延迟都个位数，5ms以内。但是终端显示去到80多ms

Image

Image

### Expected behavior

是否应该以为ping或tcping为基准？

### Actual behavior

不清楚现在具体是怎么检测的

### Logs / screenshots

_No response_

### Before submitting

binaricat:
## Summary

- measure a small bidirectional probe only after the SSH statistics channel is ready
- exclude channel opening and server-statistics execution from the displayed value
- label the result as SSH network latency in the terminal and System Overview across all supported languages
- keep Mosh and EternalTerminal wording honest: their value follows the companion SSH path
- return no latency for the buffered EternalTerminal fallback instead of reporting a misleading value
- close blocked or late probe channels cleanly on timeouts and failures
- add regression coverage for the 5ms-network / 75ms-channel-setup case and failure paths

## Root cause

Netcatty 1.1.60 started timing before opening a fresh SSH statistics channel and stopped on the first output. The displayed number therefore included channel setup and server request handling. A 5ms network path could consequently appear as roughly 80ms.

The updated path first opens the existing statistics channel. It then starts the timer, sends a one-byte probe over that channel, and stops only when the remote marker returns. This measures the bidirectional path used by the active SSH connection while excluding the expensive setup and statistics command. It does not depend on ICMP, a second TCP connection, or direct target reachability, so proxy and jump-host routes continue to use their actual SSH path.

The buffered system-SSH fallback for some EternalTerminal sessions cannot expose first-byte timing. That path now returns no latency instead of a false high value.

## Verification

- deterministic regression: 75ms channel setup plus 5ms network round trip now reports 5ms; the same test failed as 80ms before the fix
- split markers and pre-marker output are handled correctly
- timeouts, late callbacks, and probe-write failures close their channels and return a clean failure
- direct SSH, Mosh companion, direct ET companion, and buffered ET fallback tests pass
- targeted tests passed consistently across three runs
- full suite: 4667 passed, 3 skipped, 0 failed
- lint passed
- production build passed
- browser visual check confirmed the terminal shows 5ms and the expanded panel labels it SSH network latency without layout regressions

Closes #2074

### DIFF
--- a/application/i18n/locales/en/systemManager.ts
+++ b/application/i18n/locales/en/systemManager.ts
@@ -41,7 +41,7 @@ export const enSystemManagerMessages: Messages = {
   'systemManager.overview.system': 'System',
   'systemManager.overview.kernel': 'Kernel',
   'systemManager.overview.swap': 'Swap',
-  'systemManager.overview.latency': 'Network latency',
+  'systemManager.overview.latency': 'SSH network latency',
   'systemManager.overview.cpuCores': 'CPU cores',
   'systemManager.overview.disks': 'Disks',
   'systemManager.overview.interfaces': 'Network interfaces',
--- a/application/i18n/locales/en/systemManager.ts
+++ b/application/i18n/locales/en/systemManager.ts
@@ -41,7 +41,7 @@ export const enSystemManagerMessages: Messages = {
   'systemManager.overview.system': 'System',
   'systemManager.overview.kernel': 'Kernel',
   'systemManager.overview.swap': 'Swap',
-  'systemManager.overview.latency': 'Latency',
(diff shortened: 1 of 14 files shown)


binaricat:
@codex review

binaricat:
@codex review

PR #2148, opened 2026-07-12
fix(nix): restore release metadata to v1.1.63

binaricat:
## Summary
- v1.1.64 release/tag was deleted after publish.
- #2146 already merged and left `nix/release.nix` on **1.1.64** hashes that no longer exist.
- Restore metadata to **v1.1.63** (current Latest).

## Test plan
- [x] Matches pre-#2146 `nix/release.nix` content
- [ ] Merge so flake users do not fetch deleted assets

### DIFF
--- a/nix/release.nix
+++ b/nix/release.nix
@@ -1,14 +1,14 @@
 {
-  version = "1.1.64";
+  version = "1.1.63";
 
   sources = {
     x86_64-linux = {
       appImageArch = "x86_64";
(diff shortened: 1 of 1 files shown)


binaricat:
Closing as obsolete: re-publishing v1.1.64 with the compact terminal fixes. Nix metadata will be updated from the published AppImages after the release (protected-main cannot be pushed from the release job).

PR #2202, opened 2026-07-14
fix(sftp): keep sidebar path sticky across terminal focus and reconnect

### LINKED ISSUE #2193: [Feature] 多开终端且打开SFTP
### Problem / pain point

多开终端且打开SFTP ,隔段时间点击终端，目录会刷新到用户根目录下。有时候不会用到目录跟随终端打开目录。就是希望目录不要总是刷新，且保持在自己打开的地方。

Image

### Proposed solution

无

### Alternatives considered

无

### Related area

SFTP

Nice to have

### Before submitting

binaricat:
## Summary
- Keep a healthy same-host SFTP sidebar tab when switching terminal focus instead of forcing a reconnect that often lands on `$HOME`.
- Remember the last browsed path per endpoint while the panel is open, and pass it as `initialPath` when a reconnect is still required.
- On automatic SFTP reconnect, restore the previous `currentPath` instead of falling through to home after the short-lived shared host cache expires.

## Test plan
- [ ] Open SFTP on a remote host, navigate deep, disable "follow terminal directory".
- [ ] Focus another terminal on the same host — path should stay put.
- [ ] Wait >60s, then switch terminals again — path should still not jump to home.
- [ ] Simulate SFTP channel drop/reconnect — should return to the last browsed path.
- [ ] Enable follow mode and `cd` in the terminal — follow behavior still works.
- [ ] `node --import tsx --test application/state/sftp/sftpReopenLocation.test.ts components/sftp/sftpSidePanelAutoConnect.test.ts application/state/sftp/sftpConnectStartPath.test.ts components/sftp/sftpFollowTerminalCwd.test.ts`

Fixes #2193

### DIFF
--- a/application/state/sftp/sftpReopenLocation.test.ts
+++ b/application/state/sftp/sftpReopenLocation.test.ts
@@ -4,6 +4,7 @@ import assert from "node:assert/strict";
 import {
   getSftpCurrentPathMemoryKey,
   getSftpReopenMemoryKey,
+  resolveSftpAutoConnectPath,
   resolveSftpOpenLocation,
 } from "./sftpReopenLocation.ts";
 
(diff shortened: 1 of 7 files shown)


binaricat:
@codex review

binaricat:
Addressed the latest Codex notes:

- **P1 session rebind**: on terminal focus change we no longer reuse a healthy same-endpoint tab; SFTP reconnects with the new `sourceSessionId` (covers proxy/jump differences that host keys miss). Same endpoint reconnects in place with the remembered browsed path so we keep sticky dirs without stacking tabs.
- **P2 path memory**: before that rebind we snapshot + republish the visible path under the active session memory key so reopen from the newly focused terminal keeps the directory.

binaricat:
@codex review

binaricat:
Addressed the latest Codex note:

- **Promoted editor save**: same-endpoint session rebind now uses `forceNewTab` when any editor tab still references the current SFTP connection id, so Save no longer hits “SFTP connection is no longer available”.

binaricat:
@codex review

binaricat:
Addressed the latest Codex note:

- **Active transfers**: same-endpoint session rebind now uses `forceNewTab` when any pending/transferring task still references the current connection id, so in-flight upload/download/copy is not cut off by the rebind.

binaricat:
@codex review

PR #2273, opened 2026-07-17
Harden Catty terminal context handling

binaricat:
## What changed

- Bound terminal command, polling, and failure output before it enters Catty's model context, while keeping the full result available through paged local handles.
- Add per-turn read quotas, expiry, secure spill-to-disk, session cleanup, and duplicate-read protection for saved tool output.
- Make request-too-large recovery safe after terminal side effects, including exact replay of completed commands and background jobs.
- Strengthen context compaction with secret redaction, integrity repair, reusable two-pass summaries, snapshots, and compact activity restoration.
- Document the Grok CLI source study and the resulting Netcatty comparison and recommendations.

## Why

Terminal output is unbounded and can fill the context window in a single tool call. Failures, long-running monitors, repeated reads, and request retries were especially risky because they could either re-inject the same output or repeat a command with side effects. This change makes those paths bounded and recoverable without losing access to the original output.

## User impact

Catty should remain responsive during noisy terminal sessions, recover more reliably when a model request is too large, avoid repeating terminal actions during recovery, and retain useful history without exposing saved secrets to the model.

## Verification

- `node --import tsx --test infrastructure/ai/harness/*.test.ts infrastructure/ai/harness/turnDrivers/*.test.ts infrastructure/ai/*.test.ts` (329 passed)
- `node --test electron/bridges/tempDirBridge.test.cjs` (5 passed)
- `npm run lint -- --quiet`
- `npm run build`
- `git diff --check origin/main...HEAD`

The full repository suite still has one unrelated pre-existing SSH authentication retry failure; all tests covering this change pass.

### DIFF
--- a/components/ai/cattyHistoryReplay.test.ts
+++ b/components/ai/cattyHistoryReplay.test.ts
@@ -75,6 +75,22 @@ test("buildHistoricalToolResultReplayText replaces historical terminal output wi
   assert.match(replay, /command=npm run build/);
   assert.match(replay, /status=error/);
   assert.doesNotMatch(replay, /BUILD BUILD BUILD/);
+  assert.doesNotMatch(replay, /Re-run terminal_execute/);
+  assert.match(replay, /do not execute the command again/i);
+});
+
+test("buildHistoricalToolResultReplayText omits terminal poll output too", () => {
+  const replay = buildHistoricalToolResultReplayText({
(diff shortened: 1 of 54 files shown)


binaricat:
@codex review

binaricat:
@codex review

PR #2299, opened 2026-07-18
fix(terminal): recover WebGL renderer after context loss

### LINKED ISSUE #2296: [Bug] Terminal occasionally goes black after WebGL context loss (no auto-recovery)
### Operating system

macOS

### Netcatty version

main @ db18de0 (root-cause analyzed against source; observed on the macOS build)

### Affected area

SSH connection / terminal, UI / layout

### Can you reproduce it?

Sometimes

### Steps to reproduce

Probabilistic, more likely with the WebGL renderer (the `auto` default on capable machines):

1. Open several terminal tabs (each creates its own WebGL context).
2. Do one of: let the machine sleep and wake, unplug/replug an external monitor, or keep using it long enough for the GPU to reclaim/lose a WebGL context.
3. One or more terminal panes go entirely black/blank and do not repaint on their own. A keystroke or a resize sometimes brings it back.

### Expected behavior

If a terminal's WebGL context is lost (GPU sleep/resume, monitor change, context budget eviction), the pane should automatically recreate the renderer and repaint, so it never stays black without user interaction.

### Actual behavior

A lost WebGL context is only disposed, not recreated, and recreation is gated on visibility/focus events that may never fire while the pane is already visible and focused — so it stays black.

Root cause (traced in source):

1. `components/terminal/runtime/createXTermRuntime.ts` (~L503–508) — the `WebglAddon.onContextLoss` handler only `dispose()`s the addon and sets `webglLoaded = false`; it does **not** schedule a rebuild or repaint.
2. `components/terminal/useTerminalEffects.ts` — rebuild (`ensureWebglRenderer`) is driven by visibility/focus/window-shown events. If context loss happens while the pane is already visible and focused, no event fires to rebuild. A fast re-show (hidden < ~3 s, unchanged size) takes a lightweight path that skips WebGL repair/refit.
3. `components/terminal/runtime/webglRendererPolicy.ts` and `infrastructure/config/xtermPerformance.ts` — there is no GPU context budget (no cap on simultaneous WebGL contexts). With many tabs open, Chromium's ~16-context limit can evict older contexts; those visible panes go black with no re-trigger.
4. The Electron main process does not listen for `powerMonitor` suspend/resume or `screen` display changes, so sleep/wake and monitor changes don't push a recovery signal to the renderer.

Suggested fix:
- In `onContextLoss`, directly schedule a renderer rebuild + forced repaint instead of relying on later visibility events.
- In the main process, listen for `powerMonitor` `resume` and `screen` display changes and forward them to trigger `ensureWebglRenderer` + a forced repaint.
- Add a GPU context budget / graceful degradation: cap simultaneous WebGL contexts and downgrade background panes to the DOM renderer when the cap is exceeded.

Workaround: set the renderer type to DOM (non-WebGL) in settings — this removes WebGL context loss entirely.

### Logs / screenshots

Not attached (visual black-screen; intermittent). Can capture a screen recording on the next occurrence if helpful.

### Before submitting

- [X] I searched existing issues and did not find a duplicate
- [X] I removed passwords, private keys, and other secrets from this report

binaricat:
## Summary

- automatically rebuild the xterm WebGL renderer after context loss
- repaint immediately through the DOM fallback and again after the rebuild attempt
- coalesce recovery attempts and stop retrying after repeated context losses
- cancel pending recovery when the renderer is suspended or the runtime is disposed
- add regression coverage for recovery, circuit breaking, cleanup, and disabled-WebGL behavior

## Validation

- `node --test --import tsx components/terminal/runtime/webglContextLossRecovery.test.ts components/terminal/terminalTabVisibilityRecovery.test.ts`
- `npx eslint components/terminal/runtime/createXTermRuntime.ts components/terminal/runtime/webglContextLossRecovery.test.ts`
- `npx tsc --noEmit`

Closes #2296

### DIFF
--- a/components/terminal/runtime/createXTermRuntime.ts
+++ b/components/terminal/runtime/createXTermRuntime.ts
@@ -484,6 +484,11 @@ export const createXTermRuntime = (ctx: CreateXTermRuntimeContext): XTermRuntime
 
   let webglAddon: WebglAddon | null = null;
   let webglLoaded = false;
+  // Once repeated context losses trip the breaker, this runtime stays on the
+  // DOM renderer. In particular, reveal/resume calls to ensureWebglRenderer and
+  // suspend/resume cycles must not rearm WebGL. A newly-created runtime (for
+  // example after a deliberate renderer preference change) gets a fresh flag.
+  let webglCircuitBroken = false;
   let webglRecoveryTimer: ReturnType<typeof setTimeout> | null = null;
   let runtimeDisposed = false;
   const webglContextLosses: number[] = [];
@@ -515,7 +520,7 @@ export const createXTermRuntime = (ctx: CreateXTermRuntimeContext): XTermRuntime
   };
 
   const loadWebglRenderer = () => {
-    if (webglLoaded || !performanceConfig.useWebGLAddon) return;
+    if (webglLoaded || webglCircuitBroken || !performanceConfig.useWebGLAddon) return;
     let nextWebglAddon: WebglAddon | null = null;
     try {
       // WebglAddon constructor only accepts `preserveDrawingBuffer?: boolean`.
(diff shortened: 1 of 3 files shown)


binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review latest commit `94881c05`

PR #2301, opened 2026-07-18
Fix stored-key reconnect decryption race

### LINKED ISSUE #2295: [Bug] Reconnect to a stored-key host intermittently fails until the host is recreated
### Operating system

macOS

### Netcatty version

main @ db18de0 (root-cause analyzed against source; observed on the macOS build)

### Affected area

SSH connection / terminal, Host vault / keychain

### Can you reproduce it?

Sometimes

### Steps to reproduce

Intermittent (a race), but the pattern is consistent:

1. Configure a host that authenticates with an **imported/pasted SSH key** (a key whose `source` is `imported`/`generated`, i.e. its private key is stored in the vault — not a file-reference key).
2. Connect to the host — the **first** connection succeeds.
3. Close the session tab, wait a moment, then reconnect to the **same** host.
4. Occasionally the reconnect fails at the authentication stage (no server "Permission denied" — it fails before/without usable key material).
5. **Deleting the host and recreating it with the same details makes it work again.**

Suspected triggers that widen the race window (any that causes the vault to re-decrypt): cold-starting the app and connecting within the first ~1–2 s; editing/saving any host or key while a session is open; a second window writing the vault (fires a `storage` event); rapid close/reconnect loops.

### Expected behavior

Reconnecting to a host with a stored SSH key should always work, because the key material is already saved and was usable on the first connect. If the private key cannot be decrypted in time, the client should wait for/retry decryption or fall back to the on-disk identity files — not silently fail authentication.

### Actual behavior

The private key is stored encrypted (`enc:v1:` via Electron `safeStorage`) and decrypted **asynchronously** into memory. On reconnect there is a window where the key is still in encrypted form, and the connection path then discards it and disables the local fallback, causing a hard auth failure with no self-recovery. Recreating the host works because by then decryption has completed.

Root cause (traced in source):

1. `infrastructure/persistence/secureFieldAdapter.ts` — `decryptField()` returns the value **unchanged** when the credential bridge isn't ready (so `enc:v1:` can pass through); `decryptKeySecrets()` decrypts `key.privateKey`.
2. `application/state/useVaultState.ts` — `handleStorage()` (~L815–833) re-runs `decryptKeys` on `storage` events; closing sessions / edits / multi-window writes can retrigger this and leave a window where keys are still encrypted.
3. `components/terminal/runtime/createTerminalSessionStarters.ts` — `hasEncryptedPrimaryKey = isEncryptedCredentialPlaceholder(key?.privateKey)`; `allowsLocalIdentityFallback = !resolvedAuth.keyId` (**when the host has a keyId, the local `identityFilePaths` fallback is disabled**); the encrypted private key is stripped to `undefined` by `sanitizeCredentialValue` (`domain/credentials.ts`), so `hasKeyMaterial` becomes `false` → authentication fails with no recovery.

Suggested fix (either/both):
- When a `keyId` is present but the key material can't be resolved (`isEncryptedCredentialPlaceholder(key.privateKey)` and not a reference key), still allow falling back to `identityFilePaths` instead of failing.
- Don't let `decryptField`/`decryptKeySecrets` silently keep `enc:v1:` when the bridge isn't ready — await/retry decryption (or surface a clear "key could not be decrypted" error) before the connect boundary strips it.

Workaround that avoids the bug entirely: use a **file-reference key** (`source: reference`, pointing at `~/.ssh/id_*`). Reference keys are read from disk at connect time and never go through the `enc:v1:` encrypt/decrypt path.

### Logs / screenshots

Not yet captured (intermittent). Will attach an `netcatty-ssh.log` excerpt from Settings → System → SSH Debug Logs on the next occurrence; expected to show the auth stage with no usable key material rather than a server-side rejection.

### Before submitting

- [X] I searched existing issues and did not find a duplicate
- [X] I removed passwords, private keys, and other secrets from this report

binaricat:
## Summary

- wait for vault credential decryption before resolving a missing stored key during SSH and Mosh startup
- resolve reconnect authentication from the latest in-memory key and identity collections
- surface permanent vault decryption failures instead of attempting authentication without key material
- add focused tests for asynchronous vault readiness and failure propagation

## Testing

- `node --test --import tsx application/state/vaultInitStore.test.ts components/terminal/runtime/createTerminalSessionStarters.mosh.test.ts components/terminal/runtime/createTerminalSessionStarters.test.ts` (78 passed; 2 unrelated timing-sensitive output-pipeline assertions failed in the combined run and passed when rerun in isolation)
- `node --test --import tsx --test-name-pattern="session data waits|display insertion" components/terminal/runtime/createTerminalSessionStarters.test.ts`
- `npx eslint application/state/vaultInitStore.ts application/state/vaultInitStore.test.ts application/state/useVaultState.ts components/Terminal.tsx components/terminal/runtime/createTerminalSessionStarters.ts components/terminal/runtime/createTerminalSessionStarters.types.ts`
- `npx tsc --noEmit`

Closes #2295


binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review latest commit `b04618e2`

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

PR #2307, opened 2026-07-18
fix(terminal): keep input echo responsive after log floods

### LINKED ISSUE #2306: [Bug] laggy input after big log
### Operating system

Windows

### Netcatty version

1.1.69

### Affected area

SSH connection / terminal

### Can you reproduce it?

Sometimes

### Steps to reproduce

1. Open Netcatty and connect to remote server via SSH
2. make output of big log file (for example I used `cat /var/log/auth.log`
3. Try to hold down any button or type fast 
4. See the content in terminal become "laggy" and feels like there is kind of "high ping" or input lag

### Expected behavior

The input content looks fast like always

### Actual behavior

the content in terminal become "laggy" and feels like there is kind of "high ping"

### Logs / screenshots

Normal behaviour:
https://github.com/user-attachments/assets/38937aee-427d-4f69-a5ce-09a454600d36
Behaviour after cat /var/log/auth.log
https://github.com/user-attachments/assets/083f6da0-c980-4465-a785-14f019ec496c

### Before submitting

binaricat:
Closes #2306

## Summary
- after real user input is sent, briefly resume a pressure-paused terminal output stream so a trailing remote echo can arrive sooner
- cap the extra output window at 256 KiB above the existing high watermark; repeated keys cannot raise the cap
- keep background-paused sessions, automated input, protocol replies, and Ctrl+C on their existing paths
- add regression coverage for write-before-resume ordering, the hard cap, repeated input, renderer pause, and automatic input

## Live reproduction
Using the provided SSH test host, idle key echo was about 28-32 ms. During a 16 MiB LF output flood, echo took roughly 0.9-2.7 s and arrived behind about 0.4-1.6 MiB of older output, with one probe exceeding 20 s. That showed the remote echo can be held behind older SSH output. The fix gives that trailing echo a bounded chance to cross an already-paused output stream without dropping or reordering output.

## Verification
- 179 focused flow, input, output-buffer, renderer queue, and coalescer tests passed
- `npm run lint -- --quiet`
- `npm run build`
- `git diff --check`
- full `npm test` on the earlier revision: 5927 passed, 7 skipped, 1 unrelated existing SSH auth-retry test failed; the failing test also failed in isolation


binaricat:
Closing this revision after final review: the bounded 256 KiB resume does not reproduce or cover the 0.4–1.6 MiB echo ordering observed on the test host, so it cannot be claimed as a fix for #2306. The renderer-force-drain approach was also removed because it could freeze the UI. A correct follow-up needs the reporter Windows renderer/fresh-tab A/B or an equivalent Windows reproduction that fails before the change and improves after it.

PR #2348, opened 2026-07-21
[codex] Persist terminal highlights across scrolling

mam15mon:
## Summary
- render newly exposed keyword highlights synchronously before xterm paints a user scroll, including distant scrollbar jumps
- retain nearby matched-line decorations across scrollback navigation instead of deleting them at viewport boundaries
- keep persistent marker lookup constant-time across uniform scrollback trims and rebuild only for resize or structural redraws
- bound retained markers to 20 viewports and at most 1,200 matched lines so broad rules cannot create unbounded xterm trim listeners

## Root cause
The current user-scroll path intentionally waits 120 ms, scans only the settled viewport, and then deletes decorations outside that viewport. This makes unhighlighted text visible before the delayed decoration refresh. Restoring the old full-window synchronous rebuild would bring back scroll jank because it rescanned roughly five viewports on every scroll event.

## Behavior
User scrolling now reuses the contiguous indexed range and scans only newly exposed rows. A distant jump scans one viewport synchronously in xterm's `onScroll` callback, before xterm queues its own refresh. Output-driven scrolling, large-output quiet windows, recent-input protection, long-line safeguards, and alternate-buffer behavior remain unchanged.

## Validation
- `npm run lint` (0 errors; 4 pre-existing warnings outside this change)
- `node --test --import tsx components/terminal/keywordHighlight.test.ts components/terminal/keywordHighlightRegex.test.ts components/terminal/runtime/terminalOutputPressure.test.ts` (31 passed)
- `git diff upstream/main...HEAD --check`

Regression coverage includes preindexed scrolling, distant jumps, continuous jumps, scrolling during write bursts, marker disposal and reset, uniform scrollback shifts, constant-time marker lookup, bounded broad-rule retention, ANSI in-place redraws, large output, recent input, and oversized wrapped lines.

### DIFF
--- a/components/terminal/keywordHighlight.test.ts
+++ b/components/terminal/keywordHighlight.test.ts
@@ -627,6 +627,44 @@ test("persistent highlight lookup follows uniform scrollback marker shifts", () 
   }
 });
 
+test("external marker reset invalidates persistent highlight coverage", () => {
+  const raf = installAnimationFrameQueue();
+  try {
+    const {
+      term,
+      handlers,
+      getActiveDecorationCount,
+      getTranslateCount,
+      markers,
+      resetTranslateCount,
+    } = createFakeTerminal("hello DEPLOY world", { lineCount: 9 });
+    term.rows = 3;
+    const highlighter = new KeywordHighlighter(term as never);
+    const rules: KeywordHighlightRule[] = [
+      {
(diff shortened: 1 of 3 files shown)


mam15mon:
@codex review

PR #2354, opened 2026-07-21
Support key authentication in host CSV import/export

### LINKED ISSUE #2337: [Feature] CSV import/export is missing KeyPath and Passphrase columns for key-based hosts
### Problem / pain point

CSV import/export (Vault → Hosts → Import/Export CSV) only supports password-based hosts. The CSV header is fixed to `Groups, Label, Tags, Notes, Hostname/IP, Protocol, Port, Username, Password` — there's no `KeyPath` or `Passphrase` column, on either the export or the import side.

I checked the code and confirmed this is symmetric (not an export-only oversight):
- Export: `domain/vaultImport/csvExport.ts` — `exportHostsToCsv()` writes exactly those 9 columns; `host.identityFilePaths` and any key/passphrase info are never read.
- Import: `domain/vaultImport.ts` — the header-matching logic (`findHeaderIndex`) only looks for `groups/label/tags/notes/hostname/protocol/port/username/password` variants; there's no equivalent lookup for a key path or passphrase column.

So round-tripping a vault through CSV silently drops the auth method for any key-based host: the row exports with an empty `Password` field and nothing to indicate it should have used a private key, and there's no column to put a key path back on import either.

This is the same gap that was previously fixed for the MCP tools (`vault_hosts_create`/`vault_hosts_update` gained `keyPath` in #2210/#2212 and `passphrase` in #2220/#2222) — CSV import/export just hasn't been updated to match.

### Proposed solution

Add two optional columns to the CSV format:
- `KeyPath` — local private-key file path, mirroring `identityFilePaths[0]` on export and populating it on import (same field the MCP tools use).
- `Passphrase` — the saved passphrase for that key, using whatever storage mechanism `vault_hosts_create`'s `passphrase` parameter already wires up (`saveKeyPassphrase`/`defaultKeyPassphrases`) rather than inventing a new one.

Keep both columns optional/blank for password-only hosts so existing CSV files and templates remain valid.

### Alternatives considered

- Using `vault_hosts_create`/`vault_hosts_update` via the AI/MCP tools instead of CSV for key-based hosts — works today, but means CSV import/export can't be used as a complete backup/migration path when any host uses key auth, which is exactly the situation raised in #2210 originally.

binaricat:
## What changed

- add optional `KeyPath` and `Passphrase` columns to host CSV templates and exports
- restore local key authentication and saved passphrases during CSV imports
- keep existing CSV files without the new columns fully compatible
- avoid changing saved passphrases when imported rows are skipped, duplicated, ambiguous, or only partially saved
- preserve key-path values across spreadsheet editing while protecting formula-like cells

## Why

Host CSV round-trips previously omitted local key authentication entirely, so key-based hosts came back without their key path or saved passphrase.

Fixes #2337.

## Validation

- focused ESLint checks for all changed files
- 150 focused import/export, passphrase, and vault-agent tests
- production build
- `git diff --check`
- iterative independent review of correctness, compatibility, and credential handling


binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

PR #2369, opened 2026-07-21
Fix quick switcher selection and workspace creation UX

binaricat:
## What changed

- Keep the New Workspace action visible while quick-switcher results scroll.
- Restore Space as the empty-search toggle in the workspace picker while preserving spaces inside search queries.
- Give checked workspace targets a persistent background state.
- Separate keyboard selection from pointer hover so scrolling beneath a stationary pointer cannot move or freeze the keyboard selection.
- Avoid per-row state updates during pointer movement to keep mouse interaction smooth.

## Why

The quick switcher used row hover events to update the keyboard selection. When keyboard navigation scrolled the list, rows moving beneath a stationary pointer could take over the selection. The same interaction also caused frequent state updates during normal mouse movement.

## Validation

- 11 focused interaction tests pass.
- Relevant lint has no errors (one pre-existing unused icon warning remains in QuickSwitcher).
- Full production build passes.
- Verified in the running app that ArrowDown moves from CI-Build-02 to the next host without jumping or freezing.

### DIFF
--- a/components/AddToWorkspaceDialog.keyboard.test.ts
+++ b/components/AddToWorkspaceDialog.keyboard.test.ts
@@ -1,0 +1,28 @@
+import assert from 'node:assert/strict';
+import test from 'node:test';
+
+import {
+  getWorkspaceTargetRowStateClass,
+  shouldToggleWorkspaceTarget,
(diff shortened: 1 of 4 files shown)


binaricat:
@codex review

PR #2371, opened 2026-07-21
fix(sftp): make recursive directory deletion reliable

### LINKED ISSUE #2365: [Bug]
### Operating system

Windows

### Netcatty version

1.1.68

### Affected area

SFTP / file browser

### Can you reproduce it?

Always (100%)

### Steps to reproduce

sftp 删除带子目录的目录会删除失败，删除时界面会刷新下，目标目录还存在。

### Expected behavior

删除成功

### Actual behavior

未能删除

### Logs / screenshots

_No response_

### Before submitting

binaricat:
## Summary

- remove the shell-based shortcut for deleting SFTP directories
- use the SFTP client's recursive delete for ordinary interactive sessions
- retain the abort-aware recursive path for stop-sensitive operations
- add a regression case with a nested directory where the shell reports success without changing the SFTP tree

## Root cause

The directory delete path could run `rm -rf` through the SSH shell and treat exit code 0 as proof that the SFTP target was removed. The shell and SFTP channel can resolve paths in different roots, and `rm -rf` also succeeds when its target is absent. Netcatty therefore refreshed the listing as if deletion succeeded even though the directory was still present.

Closes #2365

## Validation

- regression test observed failing before the fix and passing after it
- all 58 SFTP bridge tests passed (2 live integration tests skipped because credentials were not configured)
- focused ESLint on the changed bridge and test files
- `npm run build`
- `git diff --check`

### DIFF
--- a/electron/bridges/sftpBridge.cjs
+++ b/electron/bridges/sftpBridge.cjs
@@ -337,6 +337,12 @@ const statAsync = (sftp, targetPath) =>
     sftp.stat(targetPath, (err, stats) => (err ? reject(err) : resolve(stats)));
   });
 
+const lstatAsync = (sftp, targetPath) =>
+  new Promise((resolve, reject) => {
+    const inspect = typeof sftp.lstat === "function" ? sftp.lstat.bind(sftp) : sftp.stat.bind(sftp);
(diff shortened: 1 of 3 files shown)


binaricat:
@codex review

PR #2373, opened 2026-07-21
fix: use Plug icon for External MCP top-bar toggle

binaricat:
## Summary
- Replace the External MCP top-bar `Radio` icon with `Plug`
- Keep the existing enable/disable tooltip and pressed-state behavior
- Update the source-level TopTabs coverage to lock the new glyph

## Why
`Radio` reads more like signal/broadcast, which is easy to confuse with other top-bar actions. This control is about exposing Netcatty as a local MCP bridge for external clients, so `Plug` is a clearer metaphor.

## Test plan
- [x] `node --test --import tsx components/TopTabs.test.ts`
- [ ] Open Netcatty, confirm the top-bar External MCP control uses the plug icon
- [ ] Toggle the control and confirm tooltip/pressed color still work


binaricat:
关闭：该 PR 仅把顶栏 External MCP 图标从 Radio 换成 Plug。#2374 已合并，顶栏不再单独放这个按钮，相关控制已收进快捷控制面板，因此这里不再需要合入。

PR #2382, opened 2026-07-21
fix(terminal): disable incomplete Kitty keyboard negotiation by default

zopenb:
## Summary

- stop registering the incomplete Kitty keyboard protocol handlers by default
- add an experimental, persisted opt-in under Terminal > Keyboard (applies to new sessions)
- sync the setting and cover the default-off/explicit-opt-in policy with tests

## Root cause

Netcatty currently advertises support for Kitty keyboard flag `8` (`REPORT_ALL_KEYS_AS_ESC_CODES`), but `CONTROL_KEY_CODES` only encodes Escape, Tab, Enter, and Backspace. A remote TUI can therefore enable a protocol mode that Netcatty does not fully implement.

In an SSH session running the OpenCode TUI, this capability mismatch eventually caused the remote process to stop consuming stdin and spin in `futex` / `sched_yield`. Changing the renderer to DOM did not prevent the failure, while disabling only the Kitty negotiation did.

The handlers were introduced by #924 / commit `5c5ab21b100eba155533a9a67720138b024554a6` and first shipped in `v1.1.3`.

## Reproduction evidence

| Condition | Result |
| --- | --- |
| Current Kitty negotiation enabled | stress output stopped at 610; PTY `FIONREAD=1827`; remote process around 128% CPU; syscall sampling dominated by `futex` and `sched_yield` |
| Kitty negotiation disabled only | the same 1-3000 output completed in 2m34s; `FIONREAD=0`; `TIOCOUTQ=0`; no recovery path triggered; ordinary input and Backspace remained functional |

This PR deliberately does not attempt to complete the Kitty protocol implementation. It makes the existing partial implementation experimental and opt-in so unsupported capability negotiation cannot affect existing or upgraded users. A follow-up can complete the key/event encoding before enabling it by default again.

## Tests

- `node --test --import tsx domain/terminalSettings.test.ts components/terminal/runtime/kittyKeyboardProtocol.test.ts application/syncPayload.test.ts application/i18n/locales/settingsLocales.test.ts` (92 passed)
- `node --test --import tsx application/i18n/locales/*.test.ts` (12 passed)
- ESLint on all touched files
- `npm run build`


binaricat:
@codex review

PR #2388, opened 2026-07-22
fix(sftp): follow terminal cwd on macOS/BSD remotes and first open

### LINKED ISSUE #2335: [Bug] sftp终端追随无法追随
### Operating system

Windows

### Netcatty version

1.1.70

### Affected area

SFTP / file browser

### Can you reproduce it?

Always (100%)

### Steps to reproduce

[Bug] sftp终端追随无法追随

### Expected behavior

追随终端

### Actual behavior

Image

### Logs / screenshots

_No response_

### Before submitting

tces1:
## Summary

Fixes #2335 — SFTP "follow terminal directory" did not follow.

Two root causes, both fixed:

1. **Non-Linux remotes (macOS/BSD).** The `getSessionPwd` probe read the interactive shell's cwd only via `/proc/<pid>/cwd`, which does not exist on macOS/BSD. The probe therefore fell back to the home directory (or failed outright on the `allowHomeFallback:false` path that follow uses), so the panel never tracked the terminal. Added a cross-platform `read_shell_cwd` helper: Linux keeps using `/proc`, other systems fall back to `lsof -a -p <pid> -d cwd -Fn`. The shell-discovery logic already used only `ps`, so this was the single remaining `/proc` dependency.

2. **First open stayed at home.** While the SFTP panel is closed, the per-command cwd probe does not run, so `activeTerminalCwd` can be stale (still the login home even after the terminal `cd`-ed elsewhere). On that stale value the normal follow sync saw `currentPath === terminalCwd` and did nothing, leaving the panel at home until the next `cd`. Now, when the panel first becomes visible for a connected remote with follow enabled, it forces one fresh backend probe (bypassing the stale cache) and navigates to the terminal's real cwd; it re-syncs on reopen.

## Changes

- `electron/bridges/sshBridge/sessionOps.cjs`: add `read_shell_cwd` with an `lsof` fallback for non-`/proc` systems.
- `components/SftpSidePanel.tsx`: force a fresh follow probe on first visible/connected, reset on hide.

## Test plan

- [x] Reproduced on macOS via a real sshd + a faithful single-connection exec channel: before, the probe returned home/empty for an interactive session in a non-home dir; after, it returns the correct cwd.
- [x] `node -c` syntax check + existing `sshBridge` tests pass (51/51); eslint clean.
- [x] Verified in a packaged macOS build: `cd` follow works, and the first open now lands in the terminal's current directory instead of home.
- Linux behavior unchanged (still prefers `/proc`).

Co-authored-by: Cursor Agent <cursor-agent@cursor.com>


tces1:
@codex review

tces1:
@codex review

binaricat:
@codex review

PR #2415, opened 2026-07-23
fix(terminal): stop stacking Claude Code DEC 2026 full redraws (#2291)

### LINKED ISSUE #2291: [Bug] 我在使用claude终端的时候,现实异常
### Operating system

Windows

### Netcatty version

1.1.68

### Affected area

SSH connection / terminal

### Can you reproduce it?

Often (>50%)

### Steps to reproduce

Image

如图所示,如果可视空间不足,很容易出现

### Expected behavior

正常现实

### Actual behavior

页面部分元素重复显示

### Logs / screenshots

_No response_

### Before submitting

binaricat:
## Summary
- Fix Claude Code / Codex TUI **stacked double frames** when the viewport is only slightly behind the live bottom (common with short panes and sticky-scroll lag).
- In DEC 2026 full-redraw blocks (`\x1b[H` + `\x1b[2J`), **keep cursor-home** and strip only erase-display while the user is truly reading scrollback (≥2 rows), matching Pane’s approach and avoiding xterm viewport yank without appending the next frame under the previous one.
- Re-check scroll position when a delayed `\x1b[2J` chunk arrives after a held home so returning to the live bottom still gets a real clear.

Closes #2291

## Context
Issue screenshots showed whole Claude UI blocks repeated vertically. Root cause was the sync-block filter treating any `viewportY < baseY` as “scrolled up” and stripping **both** home and clear, so the next agent frame was written without returning to the origin.

## Test plan
- [x] `node --test --import tsx components/terminal/runtime/filterSyncBlockClears.test.ts components/terminal/runtime/terminalSyncBlockFilter.test.ts components/terminal/runtime/terminalSessionAttachment.test.ts` (69 pass)
- [ ] Manual: Windows or macOS, renderer Auto, run Claude Code in a short/full terminal; confirm no stacked chat chrome when the pane is tight
- [ ] Manual: scroll up ≥2 rows into scrollback while Claude is redrawing; confirm viewport is not yanked to bottom
- [ ] Manual: switch renderer to DOM as a control if needed


binaricat:
Addressed the Codex P2 on delayed bare `\x1b[2J`:

- `appendEraseScrollbackAfterFullErases` now accepts `startInDec2026SyncBlock`
- write path seeds it from `isTerminalSyncBlockOpen(term)` **before** filtering the chunk
- regression: split `?2026h` + later `H/2J` after returning to bottom must not inject `\x1b[3J`

Tests: clearTerminalViewport + session attachment coverage for this path.

binaricat:
@codex review

Issue #2262, opened 2026-07-17
[Bug] 我在使用docker logs的时候命令行自动弹出乱码


Jinmoli:
### Operating system

Windows

### Netcatty version

1.1.68

### How did you install Netcatty?

GitHub Release (.dmg / .exe / .AppImage / .deb / .rpm / .pacman)

### Affected area

SSH connection / terminal

### Can you reproduce it?

Always (100%)

### Steps to reproduce

[root@ecs-8f96-0529534 data]# docker logs hehaidaxue-moxing

   FastAPI   Starting production server 🚀

             Searching for package file structure from directories with __init__.py files
             Importing from /code

    module   📁 app
             ├── 🐍 __init__.py
             └── 🐍 main.py

      code   Importing the FastAPI app object from the module with the following code:

             from app.main import app

       app   Using import string: app.main:app

      info   Configuration sources:
              • Module: path CLI argument
              • App name: auto-discovery

    server   Server started at http://0.0.0.0:8081
    server   Documentation at http://0.0.0.0:8081/docs

             Logs:

      INFO   Started server process [1]
      INFO   Waiting for application startup.
      INFO   Application startup complete.
      INFO   Uvicorn running on http://0.0.0.0:8081 (Press CTRL+C to quit)
Corn SM model is running...
/usr/local/lib/python3.10/site-packages/sklearn/base.py:442: InconsistentVersionWarning: Trying to unpickle estimator StandardScaler from version 1.5.1 when using version 1.7.2. This might lead to breaking code or invalid results. Use at your own risk. For more info please refer to:
https://scikit-learn.org/stable/model_persistence.html#security-maintainability-limitations
  warnings.warn(
/usr/local/lib/python3.10/pickle.py:1718: UserWarning: [23:12:53] WARNING: /workspace/src/collective/../data/../common/error_msg.h:82: If you are loading a serialized model (like pickle in Python, RDS in R) or
configuration generated by an older version of XGBoost, please export the model by calling
`Booster.save_model` from that version first, then load it back in current version. See:

    https://xgboost.readthedocs.io/en/stable/tutorials/saving_model.html

for more details about differences between saving model and serializing.

  setstate(state)
/usr/local/lib/python3.10/site-packages/sklearn/base.py:442: InconsistentVersionWarning: Trying to unpickle estimator Pipeline from version 1.5.1 when using version 1.7.2.This might lead to breaking code or invalid results. Use at your own risk. For more info please refer to:
[...]

PR #2484, opened 2026-07-25
Revert #2460 OSC color handling test

binaricat:
## Summary

- Reverts #2460.
- That PR did not leave a real docker-logs fix on main: the global OSC 10/11/12 swallow was already removed before merge, and only a private-xterm regression test remained.
- After further review of the original report (#2262) and how other terminals behave, we are not taking a Netcatty-side fix for this class of issue.

## Related

- Closes nothing by itself; follow-up context on #2262 and #2464.
- Supersedes further work on #2464.

## Validation

- Revert is a pure deletion of the #2460 test block in `terminalRuntimeMount.test.ts`.


binaricat:
@codex review

PR #2494, opened 2026-07-25
fix(sftp): pin clipboard uploads to originating tab

binaricat:
## Summary

- Pin external clipboard/folder uploads to the tab/connection captured at paste/confirm time.
- After awaits, re-resolve by **tab id** (not `getActivePane`) and assert host/endpoint unchanged.
- On session reconnect for a pinned upload, reconnect that tab and prefer its host over side-wide `lastConnectedHost`.
- Do **not** update side-wide last host when reconnecting a non-active tab.

## Why

After #2493 closes the confirm dialog immediately, users can switch SFTP tabs while multi-folder clipboard uploads are still running. Each later folder re-read `getActivePane`, so remaining items could upload to the newly active host with the old path.

## Out of scope

- Per-task external cancel controllers (follow-up PR).
- No `canCommitConnectToTab` changes (avoids forceNewTab stale-ref regression from #2479).

## Test plan

- [x] `node --test --import tsx application/state/sftp/uploadTargetPin.test.ts`
- [x] eslint on touched files
- [ ] Manual: paste multiple folders → Upload → switch to another host tab mid-transfer → remaining items stay on original host (or fail closed if tab closed)

PR #2493, opened 2026-07-25
fix(sftp): close clipboard upload dialog before transfer

binaricat:
## Summary

- Close the SFTP clipboard upload confirmation immediately when the user clicks Upload.
- Let the existing background transfer path continue without a modal overlay blocking the app.
- Keep post-upload refresh via `onUploaded` after a successful transfer.
- Extract `confirmSftpClipboardUpload` and add regression tests for close-before-await ordering.

## Why

Fixes #2478: after confirming a clipboard paste upload, the dialog stayed open until the full transfer finished, freezing both side-panel and standalone SFTP.

This replaces closed PR #2479, which ballooned into pin/reconnect work and introduced regressions. This PR is intentionally minimal: dialog handoff only.

## Test plan

- [x] `node --test --import tsx components/SftpClipboardUpload.test.ts`
- [x] `npx eslint components/sftp/SftpClipboardUploadDialog.tsx components/sftp/clipboardUpload.ts components/SftpClipboardUpload.test.ts`
- [ ] Manual: paste local files into SFTP → Upload → dialog closes immediately, transfer queue runs, UI remains interactive

PR #2495, opened 2026-07-25
fix(sftp): cancel external uploads by task id

binaricat:
## Summary

- Replace the singleton `uploadControllerRef` with a **per-task** controller map.
- Register controllers on scan/task create callbacks so cancel can target the right upload.
- `SftpTransferQueue` calls `cancelExternalUpload(task.id)` instead of a global cancel.
- Scope upload-conflict prompt cancellation to the owning controller.

## Why

After #2493, a second clipboard confirm can start while the first transfer is still running. Cancelling one queue row still called parameterless `cancelExternalUpload()`, which only held the **latest** controller and could abort the other upload.

## Out of scope

- Pinning upload targets across tab switches (see #2494).

## Test plan

- [x] `node --test --import tsx application/state/sftp/externalUploadCancel.test.ts`
- [x] eslint on touched files
- [ ] Manual: start two clipboard uploads, cancel the first in the queue — second continues

### DIFF
--- a/application/state/sftp/externalUploadCancel.test.ts
+++ b/application/state/sftp/externalUploadCancel.test.ts
@@ -1,0 +1,22 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+
+test("external upload cancellation is keyed by transfer task id", () => {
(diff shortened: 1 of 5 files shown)


binaricat:
@codex review

Issue #2386, opened 2026-07-22
[Feature] Support Cursor CLI login session (reuse subscription quota) instead of requiring CURSOR_API_KEY


itzhang89:
### Problem / pain point

The built-in "Cursor" managed agent (added in #1399) only works via the Cursor SDK
authenticated with a `CURSOR_API_KEY` (either from settings or the `CURSOR_API_KEY`
env var). In `probeCursorSdkAvailability()` (electron/bridges/aiBridge/agentDiscoveryHandlers.cjs),
the agent is skipped entirely from discovery unless an API key is present:

    const authenticated = hasEnvApiKey || hasSettingsApiKey;
    ...
    if (!cursorSdkStatus.available) continue;

This means Cursor Pro/Business/Enterprise users who already run `cursor-agent login`
locally and use the CLI's "Auto" mode (which draws from their subscription's included
agent quota) cannot use that same authenticated session inside Netcatty. To use Cursor
through Netcatty's Catty Agent panel, they're forced onto the metered API instead of
their already-paid-for subscription quota — even though `claude` and `copilot` managed
agents in the same file *do* reuse the local CLI's login (`probeClaudeAuth`,
`probeCopilotAuth`) without requiring a separate API key.

This is inconsistent with how the other managed agents behave and effectively makes
Cursor the only "pay twice" option in the AI assistant settings.

### Proposed solution

Add a local-CLI-auth detection path for Cursor, similar to what already exists for
Claude Code / Copilot, so `probeCursorSdkAvailability()` can report `authenticated: true`
based on a locally logged-in `cursor-agent` CLI session (e.g. detecting credentials under
`~/.cursor` or running something like `cursor-agent status`), without requiring
CURSOR_API_KEY to be set.

If reusing the CLI's session inside the in-process @cursor/sdk isn't feasible, an
acceptable fallback would be letting the managed agent spawn the `cursor-agent` binary
directly as a subprocess (non-interactive/print mode) the same way other CLI-backed
agents are invoked, so it inherits the user's existing login/subscription quota instead
of always going through the metered SDK+API-key path.

### Alternatives considered

- Manually running `cursor-agent` inside a Netcatty terminal tab works today and does use
  the subscription's free quota, but it bypasses Catty Agent entirely — no structured tool
  calls, no approval gating, no multi-host orchestration, no chat history integration.
- Setting CURSOR_API_KEY works but bills against the metered API instead of the included
[...]

PR #2507, opened 2026-07-26
fix(ai): harden Cursor CLI auth and Ask-mode session isolation

itzhang89:
## Summary
- Isolate Cursor CLI resume sessions by Ask vs agent so Observer → Confirm/Auto no longer sticks in read-only Ask, and skip duplicate `model_call_id` assistant stream flushes.
- Resolve Cursor CLI only via `cursor-agent` (no bare `agent` PATH fallback) to avoid collisions with other agent CLIs.
- Keep a saved Cursor API key when switching CLI login ↔ API Key modes, and update Settings copy accordingly.

## Test plan
- [ ] Settings → AI → Cursor: switch CLI login ↔ API Key repeatedly; previously saved API key still present when returning to API Key mode
- [ ] With only a non-Cursor `agent` on PATH (no `cursor-agent`): CLI login probe fails cleanly; no false Cursor detection
- [ ] With `cursor-agent` logged in: CLI mode turn works; Observer maps to Ask, Confirm/Auto starts a writable agent turn (not sticky Ask)
- [ ] Unit: `node --test electron/bridges/aiBridge/agentAuthProbes.test.cjs electron/bridges/aiBridge/sdk/cursorCliDriver.test.cjs`
- [ ] Unit: `node --test --import tsx infrastructure/ai/types.test.ts components/ai/managedAgentState.test.ts`

Related to #2386


binaricat:
@codex review

PR #2508, opened 2026-07-26
feat(terminal): support Kitty/SIXEL/iTerm inline images in terminals

### LINKED ISSUE #2488: [Feature] Support Kitty/SIXEL/iTerm inline images in SSH terminals
### Problem / pain point

Netcatty's SSH terminal currently cannot render true inline raster images produced by terminal applications.

This prevents workflows such as:

- displaying remote images with `kitten icat` or `chafa --format kitty`
- image previews in terminal file managers and TUIs
- viewing plots, diagrams, and other visual output generated by remote CLI tools

Netcatty already uses `@xterm/xterm@6.1.0-beta.220`. A matching version of `@xterm/addon-image` is available and supports Kitty graphics, SIXEL, and iTerm's Inline Image Protocol, so the protocol parser and renderer should not need to be implemented from scratch.

### Proposed solution

Integrate the matching `@xterm/addon-image` package into Netcatty's xterm runtime.

Suggested initial scope:

- Load an `ImageAddon` for terminal sessions.
- Enable Kitty graphics protocol support, particularly direct/stream transmission over SSH.
- Optionally enable SIXEL and iTerm IIP through the same addon.
- Apply conservative limits for sequence size, image pixel count, and per-session image storage.
- Dispose image resources together with the owning terminal runtime.
- Optionally expose a setting to enable or disable inline image rendering.

Example initialization:

```ts
import { ImageAddon } from "@xterm/addon-image";

const imageAddon = new ImageAddon({
  kittySupport: true,
  sixelSupport: true,
  iipSupport: true,
  pixelLimit: 4096 * 4096,
  storageLimit: 128,
});

term.loadAddon(imageAddon);
```

A basic acceptance test could be:

```bash
kitten icat --transfer-mode=stream image.png
```

Please also consider Netcatty's terminal lifecycle. Hidden terminal renderers may be hibernated and later recreated from serialized text output, while image data is held separately by ImageAddon. For an initial implementation, Netcatty could prevent full renderer hibernation after inline images have been displayed. A later implementation could explicitly preserve and restore image state.

Suggested compatibility checks:

- WebGL and DOM renderers
- split panes
- terminal and window resize
- font-size and device-pixel-ratio changes
- scrollback
- reconnect and renderer hibernation/wake
- multiple simultaneous sessions and memory limits
- existing Kitty keyboard protocol behavior

Full Kitty protocol parity does not need to block the first version. Animations, Unicode placeholders, file/shared-memory transmission, advanced relative placement, and persistence across application restarts could remain out of scope initially.

### Alternatives considered

- `chafa --format symbols` and similar tools approximate images with ANSI colors and Unicode characters. This works today but is not true raster image rendering.
- Opening images through SFTP works for individual files, but it interrupts the terminal workflow and does not support applications that emit images directly.
- Implementing Kitty Graphics Protocol directly inside Netcatty would provide more control, but would duplicate substantial parsing, rendering, storage, and security work already available in xterm's ImageAddon.
- Supporting only Kitty graphics was considered, but enabling SIXEL and iTerm IIP through the same addon may provide broader compatibility for relatively little additional integration work.

### Related area

SSH / terminal

### Before submitting

Lynricsy:
## Summary

Netcatty terminals could not render true inline raster images. This PR loads `@xterm/addon-image` (matching beta train, `0.10.0-beta.220`) in the shared xterm runtime so remote programs can draw real images — `kitten icat --transfer-mode=stream`, `chafa --format kitty`, image previews in file managers and TUIs, plots from remote CLI tools.

The issue thread flagged renderer hibernation and renderer swapping as the real cost. Both are handled here rather than deferred:

- **Renderer swaps** are handled by the addon itself and verified: it patches `IRenderService.setRenderer`, detaches its canvas layers on a swap and re-inserts them on the next render. The addon is therefore loaded *after* `term.open` but *before* the WebGL renderer is created, so the patch is in place for the deferred-WebGL path, soft-hide suspend, and context-loss rebuild.
- **Hibernation** rebuilds a tab from a text-only snapshot, which cannot carry bitmaps. Instead of adding a new persistence format, a session holding images degrades to the existing **soft-hide** mode (WebGL suspended, xterm and image layers kept alive). The gate reads the live image cache, so it lifts by itself once the cache is emptied by a terminal reset or FIFO eviction — no sticky "this tab can never hibernate again" flag.

## Type of Change

- [x] New feature

## Related Issue

Closes #2488

## Changes Made

**Domain** — `domain/terminalInlineImages.ts`
- Owns the limits (storage MB, max megapixels, sequence MB) with clamps, and converts `TerminalSettings` into addon options.
- Returns `null` when inline images are off or every protocol is disabled, so the addon is not loaded at all and the terminal keeps advertising no graphics capability to the remote side.
- `domain/models/terminal.ts`: seven new `TerminalSettings` fields, defaults, and normalization (numeric fields clamp on load, matching the `hibernate*` pattern).

**Runtime** — `components/terminal/runtime/createXTermRuntime.ts`
- Constructs the `ImageAddon` from the resolved options right after `term.open`, ahead of `createWebglRendererController`; load failures fall back to no inline images instead of breaking the terminal.
- Disposes the addon before `term.dispose()` so image storage and canvas layers never outlive their terminal.
- New `XTermRuntime.hasInlineImages()`, backed by `ImageAddon.storageUsage` rather than a flag.

**Hibernate** — `components/Terminal.tsx`
- `hibernateRuntime` takes the soft-hide branch for image sessions and never evicts another tab on their behalf.
- `canFinishHibernate` re-checks after every asynchronous step (flush, serialize).
- A consumed eviction request leaves an image session soft-hidden instead of upgrading it to full hibernate.

**Settings & i18n**
- New *Settings > Terminal > Inline Images* section: master switch, per-protocol switches (Kitty / SIXEL / iTerm IIP), image cache per terminal, maximum image size, maximum sequence size, plus a note about the hibernate interaction when hidden-tab hibernation is on.
- Translated for `en`, `zh-CN`, `zh-TW` and `ru`.
- Protocol switches are added to `SYNCABLE_TERMINAL_KEYS`; the memory limits stay local, like the other per-device renderer tuning.

**Build**
- `@xterm/addon-image` added to the `vendor-xterm` manual chunk.

### Defaults

Inline images are **on** by default with conservative limits: 128 MB image cache per terminal, 16 MP (4096 × 4096) per image, 24 MB per escape sequence. All are user-adjustable and clamped on load.

## Screenshots / Demo

Verified in a Chromium harness that bundles this branch's `resolveTerminalInlineImageAddonOptions` with real `@xterm/xterm` + `@xterm/addon-image` + `@xterm/addon-webgl`, counting non-transparent pixels on the addon's canvas layers:

| step | image layer pixels | `storageUsage` |
| --- | --- | --- |
| before any image | 0 | 0 |
| SIXEL 12 × 12 | 144 | 0.000576 |
| `+` Kitty `a=T,f=24` 4 × 4 | 160 | 0.00064 |
| `+` iTerm IIP 1 × 1 PNG | 161 | 0.000644 |
| `+` load WebGL | 161 | 0.001288 |
| `+` suspend WebGL (soft-hide) | 161 | 0.000644 |
| `+` re-create WebGL | 161 | 0.001288 |
| `+` dispose | 0 layers in DOM | — |

All three protocols draw, pixels survive every DOM ↔ WebGL transition, `getImageAtBufferCell(0, 0)` resolves, and disposal removes the layers.

## Testing

- [x] Linting passes (`npm run lint`) — 0 errors (8 pre-existing warnings in `useSftpTransfers.ts`)
- [ ] Tests pass (`npm test`) — **not run in full**. I ran the touched-area subset of the `npm test` globs (`domain/*.test.ts`, `components/terminal/*.test.ts`, `application/i18n/locales/*.test.ts`): 1495 tests, **1477 pass / 18 fail**. All 18 are the `osc7Setup` shell-integration tests and are a pre-existing baseline failure — a clean `git worktree` at `main` fails the same 18 in this environment because `fish` is not installed. Please rely on CI for the full suite.
- [x] `npx vite build` succeeds; `vendor-xterm` is 693.86 kB / 183.15 kB gzip.
- [x] Manual harness above (WebGL and DOM renderers, renderer swap, disposal).
- [ ] Not applicable: `npm run generate:capability-tools` (no capability catalog change)

New tests:
- `domain/terminalInlineImages.test.ts` — clamps, disabled resolution, unit conversion, per-protocol mapping, settings defaults.
- `components/terminal/inlineImageRendering.test.ts` — source contracts: load order relative to `term.open` and the WebGL controller, dispose-before-`term.dispose`, live-storage `hasInlineImages`, and all three hibernate gates.
- `application/i18n/locales/terminalInlineImageLocales.test.ts` — key parity across the four locales, and that translations are not English copies.

## Checklist

- [x] I have added or updated relevant documentation (README feature lists in en / zh-CN / ja-JP)

## Known limitations (deliberately out of scope, per the issue)

Animations, Unicode placeholders, file/shared-memory transmission, advanced relative placement, and persistence across application restarts are not implemented. Images are still lost if a tab is closed and reopened; within a session they now survive hidden tabs because such tabs stop at soft-hide.


binaricat:
@codex review

PR #2608, opened 2026-07-29
fix(linux): always set chrome-sandbox SUID bit in deb/rpm postinst

### LINKED ISSUE #2607: [Bug] Linux .deb: chrome-sandbox left mode 0755 (not 4755) → SUID sandbox FATAL when userns unavailable
### Operating system
Linux (Ubuntu 24.04.4 LTS)

### Netcatty version
1.1.74~dev

Built from source (`npm run pack:linux` → `.deb`), installed with `dpkg -i`.

### Affected area
Crash / app won't start

### Can you reproduce it?
Always (100%) on the affected machine.

### Steps to reproduce
1. On a Linux system where unprivileged user namespaces are unavailable to the app (e.g. Ubuntu 23.10+ with `kernel.apparmor_restrict_unprivileged_userns=1`, **or** where the bundled AppArmor profile fails to attach — see root cause), install the `.deb`.
2. Launch `netcatty` from a terminal.

### Expected behavior
App launches with a working Chromium sandbox.

### Actual behavior
App aborts immediately:

```
[FATAL:sandbox/linux/suid/client/setuid_sandbox_host.cc:166] The SUID sandbox helper
binary was found, but is not configured correctly. Rather than run without sandboxing
I'm aborting now. You need to make sure that /opt/Netcatty/chrome-sandbox is owned by
root and has mode 4755.
[ERROR:content/zygote/zygote_linux.cc:669] write: Broken pipe (32)
```

Forcing the namespace sandbox (`--disable-setuid-sandbox`) instead fails with:

```
[FATAL:content/browser/zygote_host/zygote_host_impl_linux.cc:128] No usable sandbox!
... Ubuntu 23.10+ ... disabled unprivileged user namespaces with AppArmor ...
```

So **both** sandbox mechanisms are unavailable.

### Root cause

Two independent defects combine:

1. **`chrome-sandbox` SUID bit is never set.** `scripts/linux/after-install.tpl` decides whether to `chmod 4755` vs `chmod 0755` based on a `unshare --user true` probe:

   ```bash
   if ! { [[ -L /proc/self/ns/user ]] && unshare --user true; }; then
       chmod 4755 '/opt/${sanitizedProductName}/chrome-sandbox' || true
   else
       chmod 0755 '/opt/${sanitizedProductName}/chrome-sandbox' || true
   fi
   ```

   The postinst script runs **as root** under `dpkg`. Root can always create a user namespace even when unprivileged userns is restricted, so the probe always passes and the script picks `chmod 0755`. The installed binary ends up `-rwxr-xr-x` (verified: `ls -l /opt/Netcatty/chrome-sandbox` → `0755`, not `4755`). The SUID sandbox fallback is therefore broken on any machine where the app can't actually use the userns sandbox.

2. **AppArmor profile does not attach when `/opt` is a symlink.** The intended userns path relies on the bundled profile granting `userns,`. The profile declares the exec path `/opt/Netcatty/netcatty`, but AppArmor attaches by the *resolved* path. On a host where `/opt` is a symlink (here `/opt -> /home/opt`), the real path is `/home/opt/Netcatty/netcatty`, the profile never attaches, and the userns sandbox is denied. Verified: `aa-exec -p netcatty -- unshare --user true` succeeds (profile grants userns), but the app runs `unconfined` and userns is denied.

With the userns path broken (defect 2) and the SUID bit missing (defect 1), no sandbox is usable → FATAL.

### Suggested fix

Set `chrome-sandbox` to `4755` unconditionally in `scripts/linux/after-install.tpl` — the historical electron/chrome default and a working fallback whenever the userns + AppArmor path fails:

```bash
chmod 4755 '/opt/${sanitizedProductName}/chrome-sandbox' || true
```

This keeps the AppArmor userns path intact on normal installs and guarantees a working SUID sandbox otherwise. Verified locally: after `sudo chmod 4755 /opt/Netcatty/chrome-sandbox`, the app launches cleanly.

(The `/opt`-symlink AppArmor attachment issue is a secondary, environment-specific trigger; the unconditional SUID bit sidesteps it.)

### Logs / screenshots
See Actual behavior above.

### Before submitting

chenqin231:
## Summary

On Linux, the `.deb` / `.rpm` post-install script can leave `chrome-sandbox` at mode `0755`, so on machines where the app cannot use the unprivileged user-namespace sandbox, Chromium aborts on startup:

```
[FATAL:sandbox/linux/suid/client/setuid_sandbox_host.cc:166] The SUID sandbox helper
binary was found, but is not configured correctly ... must be owned by root and have
mode 4755.
```

The post-install userns probe (`unshare --user true`) runs **as root** under `dpkg`/`rpm`. Root can create a user namespace even when unprivileged userns is restricted (e.g. Ubuntu 23.10+ with `kernel.apparmor_restrict_unprivileged_userns=1`), so the probe always passes at install time and the script picks `chmod 0755`. When the app itself then cannot use the userns sandbox, no sandbox is usable → FATAL.

This sets `chrome-sandbox` to `4755` unconditionally — the historical electron/chrome default and a safe fallback: where the bundled AppArmor profile grants userns, Chromium still prefers the namespace sandbox; elsewhere the SUID sandbox keeps the app launchable.

## Type of Change

- [x] Bug fix

## Related Issue (optional)

Closes #2607

## Changes Made

- `scripts/linux/after-install.tpl`: replace the root-context userns probe (which always passes and can leave `chrome-sandbox` at `0755`) with an unconditional `chmod 4755`, with a comment explaining why.

## Screenshots / Demo

Reproduced on Ubuntu 24.04.4 (`apparmor_restrict_unprivileged_userns=1`, `/opt` symlinked so the bundled AppArmor profile did not attach):

- Before: launching `netcatty` aborted with the SUID sandbox FATAL above; `ls -l /opt/Netcatty/chrome-sandbox` → `-rwxr-xr-x` (0755).
- After `sudo chmod 4755 /opt/Netcatty/chrome-sandbox` (the state this change produces at install time): app launches cleanly, no sandbox error.

## Testing

- [x] Tests pass (`node --test --import tsx scripts/electron-builder-config.test.cjs` → 18/18)

Note: this only changes the packaged post-install shell template, so `npm run dev` / lint / capability-spec regen are not affected.

## Checklist

### DIFF
--- a/electron-builder.config.cjs
+++ b/electron-builder.config.cjs
@@ -257,7 +257,8 @@ module.exports = {
     deb: {
         // Use gzip instead of default xz(lzma) for better compatibility with
         // Deepin OS and other distros that have issues with lzma decompression
-        compression: 'gz'
+        compression: 'gz',
+        afterInstall: 'scripts/linux/after-install.tpl'
     },
     rpm: {
         // Default fpm/electron-builder RPM compression is "xzmt" (multi-threaded
@@ -265,6 +266,7 @@ module.exports = {
         // rpmbuild fails with exit 127 during packaging. gzip is portable and
         // matches our deb preference for older distros.
         compression: 'gzip',
+        afterInstall: 'scripts/linux/after-install.tpl',
         fpm: [
             // Avoid rpm's generated /usr/lib/.build-id symlinks. Those hashes
             // are global on the host, so owning them can conflict with other RPMs.
--- a/scripts/electron-builder-config.test.cjs
+++ b/scripts/electron-builder-config.test.cjs
@@ -254,22 +254,28 @@ test("windows zip follows the requested build architecture", () => {
   );
 });
 
-test("linux FPM packages refresh the hicolor icon cache after install and remove", () => {
+test("linux FPM packages use the custom post-install template", () => {
   const fs = require("node:fs");
   const path = require("node:path");
 
-  assert.equal(
-    config.pacman.afterInstall,
-    "scripts/linux/after-install.tpl",
-    "pacman.afterInstall must point at the custom FPM post-install template",
-  );
+  for (const [target, afterInstall] of Object.entries({
+    deb: config.deb.afterInstall,
+    rpm: config.rpm.afterInstall,
+    pacman: config.pacman.afterInstall,
+  })) {
+    assert.equal(
+      afterInstall,
+      "scripts/linux/after-install.tpl",
+      `${target}.afterInstall must point at the custom FPM post-install template`,
+    );
+  }
   assert.equal(
     config.pacman.afterRemove,
     "scripts/linux/after-remove.tpl",
     "pacman.afterRemove must point at the custom FPM post-remove template",
   );
 
-  for (const relPath of [config.pacman.afterInstall, config.pacman.afterRemove]) {
+  for (const relPath of [config.deb.afterInstall, config.pacman.afterRemove]) {
     const file = path.join(__dirname, "..", relPath);
     const contents = fs.readFileSync(file, "utf8");
     assert.match(
--- a/scripts/linux/after-install.tpl
+++ b/scripts/linux/after-install.tpl
@@ -10,7 +10,7 @@ else
     ln -sf '/opt/${sanitizedProductName}/${executable}' '/usr/bin/${executable}'
 fi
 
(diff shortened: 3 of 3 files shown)


binaricat:
@codex review

binaricat:
@codex review

PR #2658, opened 2026-08-01
feat(tabs): show host name on split-workspace tabs instead of "Workspace"

chenqin231:
## Summary

A tab that holds a **split workspace** showed the generic default title **"Workspace"**, so with several tabs open you couldn't tell which host each one belonged to. This derives the tab label from the workspace's focused session instead, so the tab reads as a concrete host.

Before / after:

- Local shell split → `Localhost` (its host label) instead of `Workspace`
- SSH split → the configured host name instead of `Workspace`
- Mixed hosts in one workspace → focused host name + `+N` suffix for the rest
- A **user-renamed** workspace keeps its custom title (unchanged)

## Type of Change

- [x] New feature

## Related Issue (optional)

N/A

## Changes Made

- `domain/sessionTabTitle.ts` — add `DEFAULT_WORKSPACE_TITLE` sentinel and `resolveWorkspaceTabLabel()`, reusing `resolveSessionTabTitle` so workspace labels match exactly what single-pane tabs already show.
- `domain/workspace.ts`, `application/state/useSessionState.ts` — reference the shared `DEFAULT_WORKSPACE_TITLE` constant for the default title (single source of truth for the sentinel).
- `components/top-tabs/TopTabItems.tsx` — render the derived label (with a `title` tooltip for the full name) instead of `workspace.title`.
- `domain/sessionTabTitle.test.ts` — cover rename kept, multi-host `+N`, single host, and empty-session fallback.

The label only overrides while the workspace still carries the default title, so a renamed workspace is never touched.

## Screenshots / Demo

Split-workspace tab previously labeled `Workspace` now shows e.g. `Localhost` (local) or the SSH host's configured name.

## Testing

- [x] Linting passes (`npm run lint`) — on touched files
- [x] Tests pass (`npm test`) — new `resolveWorkspaceTabLabel` cases (13/13 in `sessionTabTitle.test.ts`); touched files typecheck clean
- [ ] Generated capability tool specs are updated when applicable

## Checklist

### DIFF
--- a/application/state/useSessionState.ts
+++ b/application/state/useSessionState.ts
@@ -626,7 +626,7 @@ export const useSessionState = ({
       
       setWorkspaceRenameTarget(prevTarget => {
         if (!prevTarget) return prevTarget;
-        setWorkspaces(prev => prev.map(w => w.id === prevTarget.id ? { ...w, title: name } : w));
+        setWorkspaces(prev => prev.map(w => w.id === prevTarget.id ? { ...w, title: name, autoTitle: false } : w));
         return null;
       });
       
@@ -652,6 +652,7 @@ export const useSessionState = ({
     // Create workspace
     const workspace = createWorkspaceFromSessionIds(sessionIds, {
       title: name,
+      autoTitle: false,
       viewMode: 'split',
     });
 
@@ -672,8 +673,12 @@ export const useSessionState = ({
     | { kind: 'local'; shellType?: TerminalSession['shellType']; shell?: string; shellArgs?: string[]; shellName?: string; shellIcon?: string }
     | { kind: 'host'; host: Host };
 
-  const createWorkspaceFromTargets = useCallback((targets: WorkspaceTarget[], name: string = DEFAULT_WORKSPACE_TITLE): string | null => {
+  const createWorkspaceFromTargets = useCallback((targets: WorkspaceTarget[], name?: string): string | null => {
     if (targets.length === 0) return null;
+    // A caller-provided name is an explicit title; without one the workspace is
+    // auto-titled and the tab derives a host-based label.
+    const explicitName = name?.trim();
+    const workspaceTitle = explicitName || DEFAULT_WORKSPACE_TITLE;
 
     const newSessions: TerminalSession[] = targets.map((target) => {
       if (target.kind === 'local') {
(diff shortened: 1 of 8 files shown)


chenqin231:
Thanks @Copilot / Codex — both point at the same root cause, fixed in `f24039d2`.

`resolveWorkspaceTabLabel` was deriving the label and the distinct-host `+N` count from `resolveSessionTabTitle`, which folds in the dynamic shell/agent title. That was wrong on two counts:
- those live titles are published only to `sessionPresentationStore`, not the plain session records passed to the tab → the label would go stale under `dynamicTabTitleMode` `agent`/`all` (Codex P2);
- per-pane dynamic titles differ on one host, so `+N` fired even when all panes shared a host (Copilot).

Fix uses `getSessionConnectionLabel` (rename / host label) — the stable host identity — for both the primary label and the distinct set. It's also exactly what the tab should convey (which host), and needs no presentation-store subscription. The `dynamicTabTitleMode` argument is dropped as now-unused. Tests + lint + typecheck green on touched files.

binaricat:
@codex review

chenqin231:
Good catch @Codex — fixed in the follow-up. The `+N` distinct-host set used the connection label (`customName || hostLabel`), so a renamed pane on the same host produced a spurious `+1`. It now counts distinct hosts by `hostLabel` alone; the focused pane still shows its rename as the primary label, only the host count ignores per-session renames. Added coverage for renamed-same-host and rename-with-multi-host. 15/15 tests + lint + typecheck green.

binaricat:
@codex review

chenqin231:
Verified and fixed. You're right — `submitSessionRename` (useSessionState.ts:585,597) rewrites both `customName` and `hostLabel`, so counting by `hostLabel` still mislabels a renamed same-host pane. Now counting distinct hosts by `session.hostId`, the only identity a rename leaves untouched: local terminals share the stable `LOCAL_TERMINAL_HOST_ID` and remote sessions carry their host config id, so same-host panes collapse to one regardless of renames. Focused pane still shows its rename as the primary label. Tests updated to carry hostId, including the rename-rewrites-both case. 15/15 + lint + typecheck green.

binaricat:
@codex review

chenqin231:
Fixed. The string sentinel was ambiguous, so I added an explicit `Workspace.autoTitle` flag. It's authoritative when present: rename, snippet-created, and named-at-creation workspaces set `autoTitle: false` and are never relabeled; only auto-titled workspaces derive a host label. Legacy workspaces without the flag fall back to the old default-title string check, so already-open default splits still get host labels while previously renamed workspaces keep their names. Covered a workspace explicitly named `Workspace` (kept) and an explicit auto-title (derived). 17/17 tests + lint + typecheck green on touched files.

binaricat:
@codex review

chenqin231:
Fixed. `restoreWorkspace()` rebuilt the workspace field-by-field and dropped `autoTitle`, so an explicitly-named workspace lost `autoTitle: false` on restore. It's now carried through with a boolean guard (so a deliberate `false` survives and legacy flagless workspaces stay flagless). Added restore coverage for the explicit-false and legacy-absent cases. 55/55 in the two suites + lint + typecheck green.

binaricat:
@codex review

Issue #2830, opened 2026-08-08
[Bug] 在终端面板中输入命令智能提示浮窗bug


setyg:
### Operating system

Windows

### Netcatty version

1.1.76

### How did you install Netcatty?

GitHub Release (.dmg / .exe / .AppImage / .deb / .rpm / .pacman)

### Affected area

SSH connection / terminal

### Can you reproduce it?

Always (100%)

### Steps to reproduce

准备一个境外服务器，命令行输入命令

<img width="506" height="295" alt="Image" src="https://github.com/user-attachments/assets/85b5cf1b-7d56-40e0-8608-a070c453efb4" />

准备一个境内服务器，命令行输入同样的命令

<img width="696" height="376" alt="Image" src="https://github.com/user-attachments/assets/f7ec1f05-269e-4988-8d73-5e7e59ee5033" />

境内服务器能够正常弹出智能提示浮窗，境外服务器不能，境外服务器还要反复删除输入，才能智能提示出来

### Expected behavior

智能提示脚本，应该不受网速影响，命令脚本是属于本地的，能够直接匹配提示出来

### Actual behavior

看上面的重现步骤

### Before submitting

- [x] I searched existing issues and did not find a duplicate
- [x] I removed passwords, private keys, and other secrets from this report

PR #2852, opened 2026-08-10
fix(terminal): restore autocomplete popup after HMR dispose and pre-echo gate

binaricat:
## Summary
- Re-arm `disposedRef` when the autocomplete mount effect re-runs so Vite Fast Refresh cleanup no longer permanently disables `fetchSuggestions` (input still scheduled, popup never appeared).
- Allow local history/fig/snippet matches during empty-echo SSH lag; keep plugins gated and still block sensitive/password prompts.
- Fall back pane visibility to the Terminal `isVisible` prop for popup/standalone mounts that never publish `paneVisibilityStore`.

## Test plan
- [ ] Cmd+R Netcatty Dev, type `docker` in SSH/local terminal — popup should appear
- [ ] Trigger HMR (edit a terminal file), type again — popup should still work without full reload
- [ ] Password / sensitive prompt should not show history popup
- [ ] `node --test --import tsx components/terminal/useTerminalAutocomplete.render.test.tsx components/terminal/terminalAutocompletePrompt.test.ts components/terminal/paneVisibilityStore.test.ts`

### DIFF
--- a/components/terminal/PromptDetector.test.ts
+++ b/components/terminal/PromptDetector.test.ts
@@ -70,8 +70,9 @@ test("uses reliable typed buffer when prompt echo has not started (IME / high-la
   assert.equal(result.prompt.userInput, "部署");
   assert.equal(result.prompt.cursorOffset, 2);
   // Pre-echo input is surfaced for alignment, but empty echo alone must not
-  // authorize history recording or any completions (password-shaped
-  // `read -s -p '$ '` prompts look identical until echo arrives).
+  // authorize history recording or third-party completion providers until echo
+  // validates the line. Local history/fig/snippet popups may still query from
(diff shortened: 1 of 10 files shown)


binaricat:
@codex review

binaricat:
Addressed Codex P1: restored the empty-echo `allowExternalProviders === false` fail-closed wait so silent shell-shaped prompts (`read -s -p '$ '`) cannot paint local history/fig/snippet suggestions. The HMR `disposedRef` re-arm (root cause of the popup never appearing) and pane-visibility fallback stay in place; partial-echo `#2830` path still uses `resolveAutocompleteQueryInput` once echo validates.

binaricat:
@codex review

PR #2877, opened 2026-08-10
fix(notes): keep MDX link preview toolbar over the link

binaricat:
## Summary

MDXEditor link preview toolbar (URL + open/edit/copy/unlink) was shifted into the notes tree when selecting a link in edit mode.

`.netcatty-mdx-editor` uses `container-type: inline-size` for narrow toolbar layout. MDX's `getSelectionRectangle` treats any `container-type` as a fixed containing block and subtracts the editor offset, but browsers do not create a fixed CB from `container-type` alone. The fixed anchor then used container-relative coords against the viewport and landed too far left.

## Type of Change

- [x] Bug fix

## Related Issue (optional)

N/A

## Changes Made

- Add `transform: translateZ(0)` on `.netcatty-mdx-editor` so the fixed containing block matches MDX coordinate math
- Guard with a unit assertion next to the existing toolbar container-type check

## Testing

- [x] Tests pass — `components/notes/InlineMarkdownEditor.test.tsx`
- Manual: edit mode, select a link near the left edge of the note body → preview bar should sit on the link, not over the notes tree

## Checklist

### DIFF
--- a/components/notes/InlineMarkdownEditor.test.tsx
+++ b/components/notes/InlineMarkdownEditor.test.tsx
@@ -189,6 +189,12 @@ test("note markdown toolbar remains usable in narrow panes", () => {
     styles,
     /\.netcatty-mdx-editor\s*\{[^}]*container-type:\s*inline-size;/s,
   );
+  // Keep fixed-containing-block in sync with MDX linkDialog coordinate math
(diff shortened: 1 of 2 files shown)


binaricat:
@codex review

PR #2897, opened 2026-08-11
feat(vault): 添加动态组路径支持用于自动化脚本

yintian710:
- 在脚本配置中添加 targetGroups 字段用于指定动态组路径
- 更新工具规范定义以支持动态组路径参数
- 实现组范围的连接脚本动态继承功能
- 更新 CLI 适配器以支持 --target-groups 参数
- 添加测试用例验证动态组路径映射和解析
- 更新国际化资源文件添加相关提示文本
- 重构脚本目标匹配逻辑分离显式主机和组路径匹配
- 实现组路径变更时的重映射和删除处理逻辑
- 更新 UI 组件支持组路径选择和显示功能
- 修改文档描述更新动态组路径使用说明

  ## Summary

  Add dynamic group targeting for scripts.

  Scripts can now target host groups without copying the groups' current host IDs into the script. Hosts added to a selected group later will automatically match the script, while hosts removed from the group will stop
  matching unless they are explicitly targeted or belong to another selected group.

  This avoids having to edit and re-save scripts whenever the membership of a target group changes.

  ## Type of Change

  - [ ] Bug fix
  - [x] New feature
  - [ ] Refactor / code cleanup
  - [ ] Documentation update
  - [ ] Build / CI change

  ## Related Issue (optional)

  N/A

  ## Changes Made

  - Added a group target option to script target selection.
  - Added a reusable group selection dialog for selecting one or more host groups.
  - Added persisted `targetGroups` metadata to scripts.
  - Resolve current group membership dynamically instead of materializing host IDs into scripts.
  - Match hosts in the selected group and all of its descendant groups.
  - Apply dynamic group targets to manual, output-triggered, and connection-triggered script execution.
  - Deduplicate scripts and hosts matched through explicit host targets, dynamic group targets, and all-host targets.
  - Stop matching a host after it leaves a selected group unless:
    - The host is explicitly targeted.
    - The host still belongs to another selected group.
    - The script targets all hosts.
  - Update stored script group paths when groups are renamed or moved.
  - Remove affected script group targets when groups are deleted.
  - Display dynamically inherited group scripts separately in host details.
  - Added `targetGroups` support to agent operations, CLI parameters, capability schemas, guidance, and generated tool specifications.
  - Ensure large CLI capability output is fully flushed before the command exits.
  - Added regression coverage for target resolution, output triggers, connection scripts, group lifecycle changes, UI rendering, agent operations, CLI mapping, and generated capability specifications.

  ## Screenshots / Demo

  Behavior overview:

  1. Open a script and choose **Select group** under its execution targets.
  2. Select one or more host groups.
  3. Hosts currently inside the selected groups match the script.
  4. Hosts added to the selected groups later automatically match without editing the script again.
  5. Hosts removed from the selected groups stop matching unless another target still applies.
  6. Selecting a parent group also matches hosts inside its descendant groups.

  ## Testing

  - [ ] Tests pass (`npm test`)
  - [x] Generated capability tool specs are updated when applicable (`npm run generate:capability-tools`)

  Validation completed:

  - Verified the feature locally with `npm run dev`.
  - No command-line errors or warnings were observed during local development testing.
  - 82 script target, connection trigger, output trigger, group lifecycle, and UI tests passed.
  - 13 CLI, capability adapter, and generated specification tests passed.
  - `git diff --check` passed.
  - Full `npm test` was executed. The remaining unrelated environment failures are caused by:
    - The locally installed dependencies missing `turndown`, which is already declared in `package.json` and `package-lock.json`.
    - A Node.js 23 module loader hook incompatibility in `TerminalLayerSupport.dragCleanup.test.ts`; the project CI uses Node.js 22.
  - The two CLI capability tests affected by truncated stdout were fixed and verified separately.

  ## Checklist


binaricat (dismissed):
Thanks for the feature. The core dynamic group matching logic looks sound, including nested groups, deduplication, connection ordering, and group rename/delete handling. I found several issues that should be addressed before merging:

1. **CLI flag is not parsed.** `electron/capabilities/adapters/cliAdapter.cjs` declares `--target-groups`, but `electron/cli/netcatty-tool-cli.cjs` never parses that flag into `opts.targetGroups`. As a result, the CLI cannot actually set dynamic group targets.

2. **Terminal autocomplete ignores group targets.** `components/terminal/autocomplete/snippetCompleter.ts` only checks `targetsAllHosts` and explicit host IDs, while `completionEngine.ts` passes only `hostId`. A script limited to a group can therefore be suggested and executed on a host outside that group.

3. **Snippet export/import drops `targetGroups`.** `domain/snippetTransfer.ts` does not include the new field in the export type, export mapping, import sanitization, or imported snippet construction. Exporting and re-importing a group-scoped script silently loses its scope; an onOutput script can become effectively unscoped.

4. **Generic snippet Agent tools are incomplete.** `vault.snippets.create/update` accept `kind=script`, but their input schemas and generated tool specs do not expose `targetGroups`. Please either add the field and regression coverage, or make those tools explicitly script-incompatible.

I verified the focused target/trigger/group-lifecycle tests (91 passed), lint, and production build. The existing remote `lint-and-test` check is also green, but these integration gaps remain.

binaricat:
@codex review

yintian710:
> @codex review

哥们你好快啊，太牛了

binaricat:
Thanks for the follow-up — the four issues from the previous review are fixed. One merge blocker remains in the group lifecycle path: group rename/move/delete persists groups, hosts, sources, and configs separately from the script `targetGroups` remap/cleanup. Deletion commits and releases the Vault lock before script targets are updated, and the Agent group operations also issue separate writes. If the later snippet write fails (for example quota exhaustion), the app exits between writes, or another window writes concurrently, the Vault can persist a renamed/deleted group while scripts still reference the old path.

Please include the snippet target remap/cleanup in the same Vault transaction as the group change, report success only after the joint commit succeeds, and add regression coverage for rename/delete failure plus a concurrent snippet update. This keeps the dynamic group link durable rather than only correct in the current render.

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

Issue #2920, opened 2026-08-12
[Bug] tmux Attach button opens a small window but does not attach


liubaicai:
### Environment

- OS: Windows 11
- Netcatty version: 1.1.78

### Steps to reproduce

1. Connect to a host that has an existing tmux session.
2. Open the tmux session list in the right sidebar.
3. Click the Attach button for a session.

### Expected behavior

Netcatty should attach to the selected tmux session normally.

### Actual behavior

The session is not attached in the current terminal. Instead, a small separate window opens and remains mostly blank.

### Screenshot

See the attached screenshot.

liubaicai:
<img width="1253" height="799" alt="Image" src="https://github.com/user-attachments/assets/c3c2887d-1f5e-4b89-aa21-e0375f99da79" />

binaricat:
已确认这个空白 tmux 弹窗与 Docker Shell/日志超时是同一个原因：独立弹窗一直等待主窗口的准备信号，因此没有真正开始连接。修复已通过 #2929 合入主线。

因此先按已完成关闭。相关改动尚未进入正式发布包；新版本发布后如果仍然出现空白窗口，欢迎直接重新打开并补充最新版本信息。感谢反馈。

PR #2929, opened 2026-08-12
fix(terminal): start command popups after vault hydration

binaricat:
## Summary

Fixes command terminal popups that stay blank and eventually report a connection timeout.

The regression affects Docker Shell, Docker Logs, and tmux Attach because all three use the same popup terminal path.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## Related Issue

Related to #2891
Related to #2912
Related to #2920

## Root Cause

Version 1.1.77 added a vault-hydration gate before terminal startup. The main window publishes that readiness through the shared snapshot store, but a terminal popup is a separate renderer: it hydrates its own vault state and never publishes into the main window's store.

As a result, the popup already has the correct host, title, keys, and startup command, while the nested terminal waits forever and never creates its backend session. The popup's independent 20-second timer then reports the generic connection timeout.

## Changes Made

- Let a popup terminal use that window's completed vault-hydration state.
- Preserve the existing shared readiness guard for the main window.
- Re-render the memoized terminal when popup readiness changes.
- Add regression tests for the separate-window startup path.

## Screenshots / Demo

No visual change. The popup now starts its existing terminal connection instead of remaining blank.

## Testing

- [x] Full `npm test` suite passed
- [x] Focused popup, terminal startup, Docker command, and SSH reuse tests passed (148 tests)
- [x] `npm run lint` passed (one pre-existing unrelated warning)
- [x] `npm run build` passed
- [x] `git diff --check` passed
- [ ] Reporter environment retest after a build is available

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my own code
- [x] I have added tests that prove the fix
- [x] New and existing unit tests pass locally with my changes
- [x] No documentation change is required
- [x] My changes generate no new warnings
- [x] No generated capability surface changed

### DIFF
--- a/components/Terminal.tsx
+++ b/components/Terminal.tsx
@@ -227,6 +227,7 @@ import type { CreateXTermRuntimeContext } from "./terminal/runtime/createXTermRu
 import { TerminalView } from "./terminal/TerminalView";
 import {
   getInitialTerminalStatus,
+  resolveTerminalVaultInitialized,
   shouldSuppressHostStartupCommandOnReconnect,
   shouldStartTerminalBackend,
 } from "./terminal/restoredSessionGate";
@@ -270,6 +271,7 @@ const TerminalComponent: React.FC<TerminalProps> = ({
   sessionId,
   workspaceId,
   restoreState,
+  vaultInitializedOverride,
   pendingInitialCwd,
   shellType,
   lastCwd,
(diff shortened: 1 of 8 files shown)


binaricat:
@codex review

PR #2995, opened 2026-08-14
fix(ai): keep composer typing off the draft notify path

binaricat:
## Summary

Typing in the AI composer still hitched after #2993: the first 1-2 characters painted, then the rest dumped after a long stall. That was not Streamdown. Every keystroke re-rendered the whole ChatInput tree, the first character synchronously created a store draft and emitted AI state, IME composition forwarded every pinyin letter to the parent, and Chromium spellcheck loaded on the first word.

The textarea is now uncontrolled. Parent draft state is committed on idle, blur, park, and send. Missing → composer-only drafts are treated as text-only identity churn so store listeners do not wake. Send still reads the live box text, so the last characters are not lost.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / code cleanup
- [ ] Documentation update
- [ ] Build / CI change

## Related Issue (optional)

Follow-up to #2993 (first-type jank). Streamdown warmup was not the remaining stall.

## Changes Made

- ChatInput keeps typed text in the DOM/`ref` instead of React state on every key.
- IME composition stays local until `compositionend`, blur, or park.
- Disable spellcheck/autocorrect/autocomplete on the composer.
- `setInputValue` no longer calls `enterScopeDraftMode` on the first character; flush does it after 120ms or on send/hide.
- Text-only draft writes skip `emitAIStateChanged` and mutation-version bumps.
- Parked commit runs only when the panel becomes hidden, not on every parked re-render.
- Submit commits the live textarea value before send/steer.

## Screenshots / Demo

N/A — input latency. First Chinese/IME and Latin typing in an empty AI composer should no longer freeze after 1-2 characters.

## Testing

- [x] Tests pass (`npm test`)

Local verification:

- `npx eslint --max-warnings 0` on the touched files
- `node --test --import tsx` for `aiDraftState`, `aiSessionsStore`, `ChatInput`, and `AIChatSidePanel.mountRetention` (59 tests)

## Checklist

### DIFF
--- a/application/state/aiDraftState.test.ts
+++ b/application/state/aiDraftState.test.ts
@@ -57,11 +57,26 @@ test("draftsByScopeEqualIgnoringComposerText ignores typing in the active scope"
 });
 
 test("first composer draft is treated as text-only identity churn", () => {
-  const created = { ...createEmptyDraft("catty"), text: "你" };
+  const empty = createEmptyDraft("catty");
+  const created = { ...empty, text: "你" };
   assert.equal(draftsByScopeEqualIgnoringAllComposerText({}, { "terminal:1": created }), true);
   assert.equal(
     draftsByScopeEqualIgnoringComposerText({}, { "terminal:1": created }, "terminal:1"),
     true,
+  );
+  assert.equal(
+    draftsByScopeEqualIgnoringAllComposerText(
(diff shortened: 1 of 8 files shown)


binaricat:
@codex review

PR #2996, opened 2026-08-14
fix(ai): clear the composer as soon as a message is sent

binaricat:
## Summary

After #2995, sending a prompt left the same text in the composer (see the leftover `看看这个机器配置怎` after the user bubble). Send waits on `aiSyncProviders` / model catalog before `clearScopeDraft`. While that is in flight, ChatInput's `useEffect(..., [onChange])` cleanup could run if `setInputValue` got a new identity and write the still-visible textarea back into the pending draft.

Clear the uncontrolled field immediately after `onSend()` starts. Commit-on-unmount now uses a ref so a new `onChange` callback cannot restore the last prompt.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / code cleanup
- [ ] Documentation update
- [ ] Build / CI change

## Related Issue (optional)

Follow-up to #2995.

## Changes Made

- `handleSubmit` calls `commitComposerText('')` after `onSend()` so the box empties during preflight.
- Replace `[onChange]` effect cleanup with an unmount-only commit through `onChangeRef`.

## Screenshots / Demo

The composer should be empty as soon as the user message appears, including while the turn is still streaming.

## Testing

- [x] Tests pass (`npm test`)

## Checklist

### DIFF
--- a/components/AIChatPanelContent.tsx
+++ b/components/AIChatPanelContent.tsx
@@ -159,6 +159,34 @@ export const AIChatPanelContent: React.FC<AIChatPanelContentProps> = ({
   parked = false,
   sending = false,
 }) => {
+  const mentionHosts = React.useMemo(
+    () => terminalSessions.map((session) => ({
+      sessionId: session.sessionId,
+      hostname: session.hostname,
+      label: session.label,
+      connected: session.connected,
+    })),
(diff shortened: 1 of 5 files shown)


binaricat:
@codex review

PR #3164, opened 2026-08-27
fix(terminal): recover the stuck IME punctuation deferral that swallows subsequent input

### LINKED ISSUE #3103: [Bug] vi 一个文件通过/xxxx 查询内容，第二次无法继续输入/xxx查询
### Operating system

Windows

### Netcatty version

1.1.80

### Affected area

SSH connection / terminal

### Can you reproduce it?

Always (100%)

### Steps to reproduce

Image

### Expected behavior

在vi一个文件的同时，可以多次通过/查询

### Actual behavior

无法多次，只能一次通过/查询 然后终端窗口会卡住

### Logs / screenshots

Image

### Before submitting

binaricat:
## Summary

Fixes `/` search in vi only working once. The first `/调用` search works fine, but after pressing Enter, pressing `/` again produces no reaction at all and the whole window appears frozen.

The cause is the deferred handling of IME punctuation. To fix the macOS Sogou issue where half-width punctuation gets converted to full-width (#2833), the keydown phase holds back ASCII punctuation like `/` and waits for the following insertText to decide whether to send the full-width character or re-send the original half-width one. That pending state originally had only three exits: an insertText commit, a keyup whose `key` exactly matches the held key, or a session reset.

Windows IMEs don't follow that order. After the IME consumes the key, the keyup often reports `Process` (keyCode 229) instead of the original `/`, and sometimes there is no keyup at all. The key comparison fails and the pending state is never cleared. Worse, all keypresses are blocked while a deferral is armed, so it's not just `/` — later keys routed through keypress (like capital letters) get swallowed too. The terminal simply stops reacting.

The freeze sequence reproduces directly against the original code:

- keydown `/` (keyCode 191) → deferral armed
- keyup reports `key: "Process"`, keyCode 229 → key mismatch, no flush, deferral stays armed
- subsequent keypresses `X`, `a`, `/` are all blocked; not a single byte reaches the PTY

The fix:

- keyup no longer requires an exact key match; any real key release ends the pending deferral (releases during composition are still left to insertText, so the half-width re-send doesn't race ahead of the full-width glyph).
- If a keydown arrives while a deferral has survived past its own keystroke (meaning the keyup was swallowed), the held-back character is flushed before the new key is processed. Repeated keydowns of the same key are treated as auto-repeat and keep the existing re-arm behavior, so no extra characters appear.
- The keypress block is narrowed from "all" to "the deferred key", matched by key or keyCode, so unrelated keys are not swallowed even when the deferral state goes stale.
- When the keyup reports a different key, the paired kitty release is encoded from the physical key recorded at defer time, so `Process`/229 is never encoded into a sequence.

## Type of Change

- [x] Bug fix

## Related Issue (optional)

Fixes #3103

## Changes Made

- `components/terminal/runtime/terminalImeTextInput.ts`: new keyup/keydown deferral-release predicates and `isModifierOnlyKey`; the keypress block now targets only the deferred key.
- `components/terminal/runtime/createXTermRuntime.ts`: the keyup flush uses the new predicate; a mismatched release is paired from the physical key recorded at defer time; keydown flushes the deferred character before handling the new key; the keypress block is passed the deferred key's keyCode.
- `components/terminal/runtime/terminalImeTextInput.test.ts`: adds Windows IME event-sequence cases — keyup turning into Process, neither keyup nor insertText arriving, auto-repeat and modifiers, and no half-width re-send mid-composition.
- Follow-up hardening from code review: only IME sentinel releases (`Process`/229/`Unidentified`/`Dead`/`Compose`) are rewritten to the deferred key; the recovery path synthesizes the deferred key's kitty release so no press is left stranded; modified keydowns (Ctrl/Alt/Meta) discard a stale deferral instead of injecting the lost keystroke in front of an interrupt like Ctrl+C; keyup handling is a total case analysis (sentinel takes over / own identity / unrelated keeps its identity plus a synthesized release).

The #2833 behavior is untouched: the insertText commit, the composition path, and the English-punctuation keyup fallback all fire at the same points, and their tests keep passing.

## Screenshots / Demo

No UI changes.

## Testing

- [x] Tests pass (`npm test`)

- `node --test --import tsx components/terminal/runtime/*.test.ts`: 802 cases all pass, including the new IME cases (each new case was red before its fix).
- The remaining cases under `components/terminal`: 806 all pass.
- The full `npm test` has 9 failures locally, all in the `packages/plugin-cli` packing cases and `electron/bridges/sshAlgorithms.test.cjs`; none of those files were touched in this PR. The failures come from the worktree not having its own node_modules and from local zip-validation behavior differences — unrelated to this change.

## Checklist

### DIFF
--- a/components/terminal/runtime/createXTermRuntime.ts
+++ b/components/terminal/runtime/createXTermRuntime.ts
@@ -110,6 +110,8 @@ import {
   shouldBlockKeyPressForImeTextInput,
   shouldCommitDeferredImeTextInput,
   shouldDeferKeyDownForImeTextInput,
+  shouldFlushDeferredImeTextInputOnKeyUp,
+  shouldFlushStaleDeferredImeTextInput,
 } from "./terminalImeTextInput";
 import { formatSerialLocalEcho } from "./serialLocalEcho";
 import { mapTerminalBackspaceInput } from "./terminalBackspaceInput";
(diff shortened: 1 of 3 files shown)


binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

PR #3168, opened 2026-08-27
fix(terminal): send IME commits and short pastes per character

### LINKED ISSUE #3077: [Feature] 连接奇安信堡垒机，无法进行中文输入
### Problem / pain point

Image

用Netcatty连接奇安信堡垒机的时候，搜索关键字的时候，无法键入中文，键入后不显示，按回车就是取消搜索

### Proposed solution

能否支持奇安信堡垒机中文输入

### Alternatives considered

None

### Related area

SSH / terminal

### Before submitting

binaricat:
## Summary

Fixes #3077. On QAX (奇安信) bastion hosts, an IME commit of two or more hanzi and a pasted Chinese string both vanish at the bastion's "输入搜索关键字" prompt, while typing the same characters one at a time works. This PR splits multi-character plain text at the send boundary so those payloads leave the renderer as per-character writes.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / code cleanup
- [ ] Documentation update
- [ ] Build / CI change

## Related Issue (optional)

Fixes #3077

## Changes Made

Diagnosis: the bastion treats one SSH channel write as a single keystroke and drops any multi-character chunk entirely - not even the first character shows. Single hanzi typed one at a time works, an IME commit of 2+ characters in one write shows nothing, pasting Chinese shows nothing, encoding (UTF-8 vs GB18030) is irrelevant, and normal server shells are unaffected. Xshell works there because it sends one write per keystroke, which matches the reporter's own suggestion of sending the text "逐个接收、然后逐个发给堡垒机命令行". Netcatty sent an IME commit as one write: xterm `onData` → `handleTerminalInputData` → `terminalBackend.writeToSession` → IPC → one `stream.write` of the whole string in the terminal bridge. I verified the bridge has no coalescing queue for interactive input: each non-automated IPC write maps to exactly one channel write, only ordered by the interceptor promise chain, so a renderer-side split reaches the wire as separate writes.

What the change does, renderer-side only (no new settings, no UI, no main-process changes):

- New `components/terminal/runtime/terminalPerCharacterInput.ts` holds the decision logic. `handleTerminalInputData` gained a `perCharacterWrites` option and its write site now iterates `getTextInputWireChunks(outData, ...)` - one chunk per Unicode code point, never per UTF-16 code unit, so surrogate pairs stay intact. Every chunk carries the same `{ sensitive }` options as the original single write.
- IME commits: the non-kitty literal-glyph paths (`commitImeTextInput` fallbacks and the committed-composition fallback in `term.onData`) split whenever the committed text has more than one code point.
- Short raw pastes: a raw `onData` payload with no escape sequence at all and at most 32 characters goes out per character; longer raw pastes keep today's single write. 32 characters covers the second reported failure (pasting a host name or a few hanzi into the search prompt) without degrading bulk pastes.
- Escape sequences are never split: anything containing `\x1b` keeps the single-write behavior, which leaves bracketed paste (`\x1b[200~`), negotiated Kitty CSI-u associated text, forwarded Kitty key sequences, and terminal reports exactly as they are today. Bastions don't negotiate Kitty, so those paths are untouched.
- Bookkeeping still sees the original payload exactly once: command buffer, script recorder, autocomplete, sudo autofill, scroll suppression, and broadcast all keep operating on the unsplit string. Broadcast peers still receive the original string once instead of N characters.

## Screenshots / Demo

No UI change. Behavior: on a strict bastion prompt, an IME commit of "中国" now reaches the wire as two writes (`中`, then `国`) instead of one write of "中国", and a pasted short hostname or Chinese string goes out character by character.

## Testing

- [x] Tests pass (`npm test`)

Note on the two unchecked boxes: end-to-end verification against the actual QAX bastion isn't possible for us - we cannot reach the reporter's host, and the issue thread shows the owner hit the same wall. The approach is the one the reporter proposed, so this PR needs their confirmation on that bastion. `npm test` results are identical to `main` apart from the new tests: the 8 failures in `packages/plugin-cli` packing/archive tests and `electron/bridges/sshAlgorithms.test.cjs` reproduce on a clean checkout of `main` in this environment and are unrelated.

New tests: `terminalPerCharacterInput.test.ts` covers the splitter (surrogate pairs stay on one write), the IME rule, the raw-paste cap at 32 characters, and the rule that anything containing an escape sequence never splits. `createXTermRuntime.test.ts` adds wiring assertions that the write loop reuses the same options, that the Kitty CSI-u and forwarded-key paths are not flagged for splitting, and that bookkeeping/broadcast keep the unsplit payload. Existing tests that pinned the exact single-write call shape (`terminalInterceptorInputSafety.test.ts`, `terminalImeTextInput.test.ts`, `shiftEnterText.test.ts`) were updated to the chunked form they now assert.

Remaining known path: the compose bar / snippet runner sends programmatic input through the `automated` write path with its own line-delay machinery, so a single long line from there is still one write. I left it alone since it is a different surface than the reported one, but it is the next place to apply the same split if the reporter still sees this when sending via the compose bar.

## Checklist

### DIFF
--- a/components/terminal/runtime/createXTermRuntime.test.ts
+++ b/components/terminal/runtime/createXTermRuntime.test.ts
@@ -966,3 +966,52 @@ test("alternate-screen history preview falls back to the captured session output
   assert.match(source, /outputHistory\.getPreviewRows\(\{/);
   assert.match(source, /nextOutputHistoryPreviewTop\(/);
 });
+
+test("multi-character plain text goes out as per-character writes (#3077)", async () => {
+  const { readFileSync } = await import("node:fs");
+  const source = readFileSync(new URL("./createXTermRuntime.ts", import.meta.url), "utf8");
+
+  // Strict bastion prompts (QAX) treat one channel write as one keystroke and
+  // drop multi-character chunks, so IME commits and short raw pastes must be
+  // sent one character per write instead of one write per commit.
+  assert.match(source, /perCharacterWrites\?: boolean/);
+  assert.match(
+    source,
+    /getTextInputWireChunks\(outData, options\?\.perCharacterWrites === true\)/,
+  );
+  // Every chunk carries the same write options as the original single write.
+  const writeLoopIdx = source.indexOf("for (const chunk of getTextInputWireChunks(outData");
+  assert.ok(writeLoopIdx >= 0);
+  const writeLoop = source.slice(writeLoopIdx, writeLoopIdx + 320);
+  assert.match(writeLoop, /for \(const chunk of/);
+  assert.match(writeLoop, /writeToSession\(id, chunk, \{ sensitive \}\)/);
+
+  // IME commits split the committed glyph, not the Kitty CSI-u encoding.
+  assert.match(
+    source,
+    /handleTerminalInputData\(text, \{ perCharacterWrites: shouldSplitImeTextInputForWire\(text\) \}\)/,
+  );
+  // Raw onData (unbracketed paste / committed composition) uses the capped
+  // raw-paste rule; the composition fallback uses the uncapped IME rule.
+  assert.match(
+    source,
+    /handleTerminalInputData\(data, \{ perCharacterWrites: shouldSplitRawPasteInputForWire\(data\) \}\)/,
+  );
+  assert.match(
+    source,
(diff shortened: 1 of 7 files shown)


binaricat:
@codex review

PR #3207, opened 2026-08-29
fix(terminal): guard the terminal worker against process-level errors

### LINKED ISSUE #3206: [Bug] One stray async error in the terminal worker disconnects every session at once
### Operating system

Windows

### Netcatty version

1.1.82

### Affected area

SSH connection / terminal, Port forwarding, Crash / app won't start

### Can you reproduce it?

Sometimes

### Steps to reproduce

1. Open several sessions at once (SSH tabs, an SFTP panel, a port-forward tunnel).
2. Leave them running for a while under normal use — typing, resizing, a laptop sleep/wake, a Wi-Fi/VPN change, a jump-host hop that drops.
3. Wait for any single one of those sessions to hit a stray asynchronous error.

I cannot give a deterministic trigger, and that is part of the report: the failure is
whatever async error happens to reach the worker's event loop first. See "Actual behavior"
for why the specific exception is currently impossible to capture from a normal install.

### Expected behavior

One failing session — a dropped jump host, a reset socket, a broken tunnel — should
fail on its own and leave every other session connected.

This is the project's own stated policy. `electron/bridges/processErrorGuards.cjs` opens with:

```js
// Any error with an ssh2 `level` property is a connection/auth-level error,
// never a reason to kill the entire multi-session app.
```

### Actual behavior

**Every session disconnects simultaneously**, each reporting:

```
Terminal worker exited with code 1
```

**Root cause.** All terminal, SSH, SFTP, and port-forwarding sessions live in a single
`utilityProcess` (`terminalWorkerManager.cjs`, `ensureStarted()` →
`utilityProcess.fork(workerScriptPath, ...)`). That worker installs **no**
`uncaughtException` and **no** `unhandledRejection` handler:
`electron/terminalWorker/process.cjs` `main()` never calls into
`processErrorGuards.cjs`. So Node's default behavior applies — any single unhandled async
error exits the process with code 1, and `handleExit()` then closes *every* tracked
session and emits `netcatty:exit` to each one.

The main process installs those guards (`electron/main.cjs`), and the test file covering
them is even named `mainProcessErrorGuards.test.cjs`. The protection was written to stop
exactly this failure mode; when sessions moved into the worker, the worker became the
process that has to survive them, and it is the one process without the guard.

Concrete unguarded paths I found while reading 1.1.82 (all still present on `main`):

| Location | Gap |
|---|---|
| `terminalWorker/runtime.cjs` `handleMessage()` | no `try/catch` around the whole dispatch |
| `terminalWorker/runtime.cjs` `handleSend()` | calls `listener(...)` unguarded — this is the `netcatty:write` / `resize` / `flow` / `close` / `interrupt` path. `handleRequest()` *is* wrapped, so only the fire-and-forget listeners are exposed |
| `bridges/terminalBridge.cjs` `writeToSession()` | `setTimeout(sendChunk, ...)` — a throw in a timer callback cannot be caught by any surrounding `try/catch` |
| `bridges/sshBridge.cjs` (jump-host chain) | `conn.once('error', ...)` only. After `ready`, `settled` swallows the first error and the `once` listener is then gone; a second `error` on that client hits an EventEmitter with no listener and throws |

**Why there is no log to attach.** `handleExit()` never calls `crashLogBridge`, and the
only `onWorkerExit` listener is port forwarding's tunnel-status cleanup. So a worker crash
leaves **no crash-log entry at all**. The worker's stderr goes to the main process's
inherited stderr, which is discarded under the Windows GUI subsystem. My `crash-logs/`
folder contains only `render-process-gone` entries (`exitCode=1073807364`,
`activeSessionCount: 0`) — unrelated to this. The cause is printed and then thrown away,
which is why this is hard to report properly.

Setting `NETCATTY_TERMINAL_WORKER=0` avoids the mass disconnect, which is consistent with
the diagnosis: that path registers the bridges in the main process, where the guards exist.

I have opened a PR that installs the same guards in the worker and routes worker crashes
into the crash log so future reports can carry the actual exception.

### Logs / screenshots

```
{"timestamp":"2026-08-27T13:20:10.102Z","source":"render-process-gone",
 "message":"Renderer process gone: reason=killed, exitCode=1073807364",
 "version":"1.1.81","electronVersion":"42.3.3","osVersion":"10.0.19045",
 "activeSessionCount":0,"uptimeSeconds":587}
```

Included to show what the crash log does **not** contain: there is no worker-exit record,
because nothing writes one.

### Before submitting

zxhggg:
## Summary

Every terminal, SSH, SFTP, and port-forwarding session lives in one
`utilityProcess`, and that process installs no `uncaughtException` /
`unhandledRejection` handler. So a single stray async error anywhere in the worker
exits it with code 1, and `terminalWorkerManager.handleExit()` then closes **every**
live session at once with `Terminal worker exited with code 1`.

The main process already installs these guards for exactly this reason —
`processErrorGuards.cjs` opens by saying an ssh2 connection-level error is
"never a reason to kill the entire multi-session app", and the test file covering it
is even named `mainProcessErrorGuards.test.cjs`. When sessions moved into the worker,
the worker became the process that has to survive them, and it is the one process
without the guard.

This PR installs the same guards there, and makes worker crashes visible in the
crash log so future reports can carry the cause rather than only the symptom.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / code cleanup
- [ ] Documentation update
- [ ] Build / CI change

## Related Issue (optional)

Closes #3206

## Changes Made

**`electron/terminalWorker/process.cjs`** — new `installWorkerProcessErrorGuards()`,
called at the top of `main()`. It reuses `processErrorGuards.cjs` unchanged and calls
`completeMainWindowStartup({ windowShown: true })` immediately: the worker has no window
lifecycle, so without arming runtime protection `classifyProcessError()` would read every
runtime error as a pre-startup failure and still let the process die. Suppressed errors
are forwarded to the parent as a `worker-process-error` message.

**`electron/bridges/terminalWorkerManager.cjs`** — handle that message kind and fan it out
through a new `onWorkerProcessError()` subscription, following the existing
`onTerminalInterceptorWarning()` pattern (listener set, dispatch branch, dispose clear,
export).

**`electron/main/registerBridges.cjs`** — write both worker exits and suppressed worker
errors to `crashLogBridge`. Previously `handleExit()` never reached the crash log and the
only `onWorkerExit` listener was port forwarding's tunnel cleanup, so a crash that dropped
every session left no record at all; the worker's stderr goes to the main process's
inherited stderr, which is discarded under the Windows GUI subsystem.

**Tests** — 5 cases in `process.test.cjs` (ssh2 `level` error, generic runtime error and
rejection, `ECONNRESET` reporting, `EPIPE` ignored without a report, uninstall) and 1 in
`terminalWorkerManager.test.cjs` (fan-out + dispose).

### Not fixed here

While tracing this I found specific unguarded throw sites that the guard now catches but
that are worth fixing at the source separately, so this PR stays reviewable:

- `terminalWorker/runtime.cjs` — `handleMessage()` and `handleSend()` have no `try/catch`.
  `handleRequest()` does, so only the fire-and-forget `ipcMain.on` path (`netcatty:write`,
  `resize`, `flow`, `close`, `interrupt`) is exposed.
- `bridges/terminalBridge.cjs` `writeToSession()` — `setTimeout(sendChunk, ...)`; a throw in
  a timer callback is uncatchable by any surrounding `try/catch`.
- `bridges/sshBridge.cjs` jump-host chain — `conn.once('error', ...)` only. After `ready`,
  `settled` swallows the first error and the `once` listener is then gone, so a second
  `error` on that client reaches an EventEmitter with no listener and throws.

## Screenshots / Demo

No UI change. Behavioral change: a stray async error in the worker is now logged and
suppressed instead of terminating the process and disconnecting every session.

## Testing

- [ ] Linting passes (`npm run lint`)
- [ ] Tests pass (`npm test`)
- [x] Generated capability tool specs are updated when applicable (`npm run generate:capability-tools`) — not applicable, no capability catalog change

**Please read the three unchecked boxes — I want to be precise about what I did and did not run.**

I could not run `npm ci` in my environment, so `npm run dev`, `npm run lint`, and the full
`npm test` did not run. What I did run, with `node --test` directly on the suites that have
no native/third-party dependencies:

```
electron/terminalWorker/process.test.cjs
electron/terminalWorker/runtime.test.cjs
electron/bridges/mainProcessErrorGuards.test.cjs
electron/bridges/terminalWorkerManager.test.cjs
electron/bridges/terminalWorkerManager.transferFanout.test.cjs
                                        -> 165 tests, 163 pass, 2 fail
```

The 2 failures are `Cannot find module 'ssh2-sftp-client'` and
`Cannot find module 'ajv/dist/2020'`. I confirmed they are environmental by stashing this
branch and re-running on clean `main`: **the same 2 fail identically there**. All 6 new
tests pass, and `node --check` is clean on all three changed source files.

CI runs the real suite — please treat that as the authoritative result over my partial run.

## Checklist

- [ ] I have added or updated relevant documentation — no user-facing docs affected; rationale is in code comments

### DIFF
--- a/electron/bridges/terminalWorkerManager.cjs
+++ b/electron/bridges/terminalWorkerManager.cjs
@@ -414,6 +414,7 @@ function createTerminalWorkerManager(options = {}) {
   const urgentInputPorts = new Map();
   const outputTaps = new Set();
   const terminalInterceptorWarningListeners = new Set();
+  const workerProcessErrorListeners = new Set();
   const sessionOwnedListeners = new Set();
   const sessionClosedListeners = new Set();
   const workerExitListeners = new Set();
@@ -1432,6 +1433,14 @@ function createTerminalWorkerManager(options = {}) {
       }
       return;
     }
+    if (message.kind === "worker-process-error") {
+      // The worker survived this one (its process guards suppressed it), but
+      // main still owns the crash log, so record it there for bug reports.
+      for (const listener of [...workerProcessErrorListeners]) {
+        try { listener(message); } catch {}
+      }
+      return;
+    }
     if (message.kind === "renderer-event") {
       for (const listener of [...workerRendererEventListeners]) {
         try { listener(message); } catch {}
@@ -2003,6 +2012,12 @@ function createTerminalWorkerManager(options = {}) {
     return Object.freeze({ dispose: () => terminalInterceptorWarningListeners.delete(listener) });
   }
 
+  function onWorkerProcessError(listener) {
+    if (typeof listener !== "function") throw new TypeError("Worker process error listener is required");
+    workerProcessErrorListeners.add(listener);
+    return Object.freeze({ dispose: () => workerProcessErrorListeners.delete(listener) });
+  }
+
   function onSessionOwned(listener) {
     if (typeof listener !== "function") throw new TypeError("Terminal session owner listener is required");
     sessionOwnedListeners.add(listener);
@@ -2055,6 +2070,7 @@ function createTerminalWorkerManager(options = {}) {
       supersedingSessionGenerations.clear();
       closeAllUrgentInputPorts();
       terminalInterceptorWarningListeners.clear();
+      workerProcessErrorListeners.clear();
       sessionOwnedListeners.clear();
       sessionClosedListeners.clear();
       workerExitListeners.clear();
@@ -2088,6 +2104,7 @@ function createTerminalWorkerManager(options = {}) {
     attachTerminalInterceptor,
     detachTerminalInterceptor,
     onTerminalInterceptorWarning,
+    onWorkerProcessError,
     onSessionOwned,
     onSessionClosed,
     onWorkerExit,
--- a/electron/bridges/terminalWorkerManager.test.cjs
+++ b/electron/bridges/terminalWorkerManager.test.cjs
@@ -4944,3 +4944,33 @@ test("worker exit notifies host lifecycle listeners for every active session", a
     { sessionId: "local-2", reason: "worker-exit" },
   ]);
 });
+
+test("worker process error reports are fanned out to main-process listeners", () => {
+  const child = new FakeChild();
+  const manager = createTerminalWorkerManager({
+    utilityProcess: {
+      fork() {
+        return child;
+      },
+    },
+    workerScriptPath: "/worker.cjs",
+  });
+  const seen = [];
+  const subscription = manager.onWorkerProcessError((report) => seen.push(report));
+  void manager.request("netcatty:test", {});
+
+  child.emit("message", {
+    kind: "worker-process-error",
+    source: "uncaughtException",
+    message: "Keepalive timeout",
+    level: "client-timeout",
+  });
+
+  assert.equal(seen.length, 1);
+  assert.equal(seen[0].source, "uncaughtException");
+  assert.equal(seen[0].message, "Keepalive timeout");
+
+  subscription.dispose();
+  child.emit("message", { kind: "worker-process-error", message: "second" });
+  assert.equal(seen.length, 1);
+});
--- a/electron/main/registerBridges.cjs
+++ b/electron/main/registerBridges.cjs
@@ -393,6 +393,25 @@ function createBridgeRegistrar(context) {
       : null;
     registerTransportIdleTtlSettingsSync(ipcMain, terminalWorkerManager);
     if (terminalWorkerManager) {
+      // A worker crash silently drops every live session at once. Record it in
+      // the crash log so bug reports carry the cause instead of only the
+      // "Terminal worker exited with code 1" symptom.
+      terminalWorkerManager.onWorkerExit?.((error) => {
+        try {
+          crashLogBridge.captureError("terminal-worker-exit", error);
+        } catch { /* never throw from crash logging */ }
+      });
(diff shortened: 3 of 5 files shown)


binaricat:
Closing as superseded by #3208, which was merged on August 30 and already covers issue #3206. Thank you for identifying and documenting the shared-worker failure and missing crash records.

Two independent read-only reviews compared this PR's unchanged head (`b3b7b046c06f08cc143e484c8d3fe8c903b56153`) with current main (`81ec019892288be2f2f0ee7c78293abbb7bc0758`). No necessary unique fix remains here: main contains runtime error protection, worker-error and exit logging, startup-failure handling, and send-listener isolation. The existing P1 review is valid in this PR: arming protection before initialization can swallow a startup failure and leave requests waiting. #3208 already handles that correctly, so merging this duplicate would add no needed fix.

Fresh validation on current main: 146 focused tests passed. A real Electron utility-process experiment confirmed an unprotected process exits after an injected asynchronous exception; the current guard survives an exception and rejected promise and answers the next request; an error during startup still terminates the process. This is controlled fault injection, not confirmation from the original Windows reporter or a sleep/VPN reproduction.

No source changes were made and this duplicate was not merged. The applicable fix remains #3208.

Issue #2851, opened 2026-08-10
[Bug] 合并终端后会话session丢失


Damomo11:
### Operating system

Windows

### Netcatty version

1.1.76

### How did you install Netcatty?

GitHub Release (.dmg / .exe / .AppImage / .deb / .rpm / .pacman)

### Affected area

AI assistant

### Can you reproduce it?

Always (100%)

### Steps to reproduce

1. 在A终端里使用ai会话
2. 将B终端和A终端合并
3. ai会话不是原来那个，历史会话也找不到

### Expected behavior

能继续使用原来那个ai会话

### Actual behavior

ai会话显示的不是原来那个，也不是B用过的ai会话，是其他终端用过的ai会话session。   只有在分离开终端，才能看到我需要的那个ai会话。

### Before submitting

- [x] I searched existing issues and did not find a duplicate
- [x] I removed passwords, private keys, and other secrets from this report

PR #3225, opened 2026-08-31
fix(ai): keep resumed terminal chats visible after merge

binaricat:
## Summary

Keep an AI conversation visible when a terminal resumes it from history and is then merged into a workspace.

This is a confirmed reproduction of the disappearance described in #3224, with one important condition: the chat was created on an older terminal and resumed on a current member terminal. The issue has no recording/logs establishing whether the reporter used that condition, so this PR does not claim to explain every possible recurrence or automatically close the report.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / code cleanup
- [ ] Documentation update
- [ ] Build / CI change

## Related Issue (optional)

Related to #3224. Follow-up to #2851 / #2855.

## Changes Made

- Include AI chats selected by **current workspace member terminals** in workspace history, even when the stored terminal target belongs to a previous connection.
- Preserve existing member-target matching and history ordering; do not widen matching to every chat from the same host or to chats selected by nonmember terminals.
- Include member selections in the history cache key, while retaining cache reuse for unrelated terminal changes.
- Use the same membership inputs for panel rendering and memoized active-chat tracking, so continued responses update the visible chat.
- Add history/cache and rendered-panel regressions. No chat IDs, stored scopes, messages, or external-agent continuation IDs are rewritten by this change.

### Root cause and evidence

The original merge fix is still present, including in v1.1.82. It recognizes member chats through `session.scope.targetId`. Resuming history on terminal A updates A's active selection but keeps the chat's original target. After merging B into A, the workspace history filter therefore rejects the resumed chat. The handoff selects the correct chat ID, but view normalization subsequently treats that ID as absent, demotes the panel to draft, and clears the workspace selection. Without an explicit view, an unrelated old workspace chat can be selected instead. The messages remain stored and become visible again when returning to A.

Discriminating checks: a fresh chat created on A passes; the same selected chat with an older stored terminal target fails. The handoff helper returns the correct ID in both cases. Changing only history membership restores display. Existing 50 scope/handoff tests passed before the fix; three new regressions failed before it and pass after it.

## Screenshots / Demo

Exercised the production `AISidePanelStateRoot` / `AIStateMaintenanceHost` / `AIChatPanelsHost` / `AIChatSidePanel` in a local browser fixture under StrictMode, using synthetic chats and terminal/workspace inputs:

1. Resume **A history conversation** through the real history UI.
2. Merge B into A: before the fix, the message disappears and history contains only an old workspace conversation; after the fix, the same message and selected chat remain.
3. Confirm history reselect, follow-up typing, hide/show, detach/remerge, and merge while the panel is hidden.
4. Confirm explicit **New Chat** still opens a blank draft, and another live terminal's same-host chat does not appear.

This fixture exercises the real AI state maintenance and chat UI, not the full terminal drag gesture, native Windows executable, SSH backend, or a live external-agent request. A fixture hot-update caused a duplicate-createRoot warning; fresh-page interaction did not introduce runtime errors.

## Testing

- [x] Tested locally with the production AI panel in a Vite browser fixture (details above)
- [ ] All tests pass (`npm test`) — 11,013 passed, 4 failed, 13 skipped. All four failures are unchanged `components/terminal/osc7Setup.test.ts` legacy-shell migration assertions, reproduced on a separate clean `origin/main` checkout at `e8c291373b4402b47a5f49fc895a9469f3c2d865` (37 passed / the same 4 failed). Focused AI suite: 71 passed.
- [x] Production build passes (`npm run build`)
- [x] Generated capability tool specs are unchanged (not applicable)
- [x] Fresh-page browser interaction checked for runtime errors (fixture hot-update caveat above)

Focused command:

```sh
node --test --import tsx components/ai/scopedHistorySessions.test.ts components/AIChatSidePanel.mountRetention.test.ts components/ai/sessionScopeMatch.test.ts domain/aiWorkspaceScopeInherit.test.ts domain/workspaceAiScopeHandoff.test.ts components/ai/aiPanelViewState.test.ts
```

Full test run used Node 24.11.1 after restoring Electron 42.3.3 locally. An initial Node 26 run was interrupted because the missing Electron runtime caused repeated installer attempts; it is not used as product-failure evidence. Build emits existing chunk-size/dynamic-import warnings; lint has no findings.

## Checklist

- [x] Relevant behavior and reproduction are documented above and in regression tests
- [x] No breaking changes or storage migration

### DIFF
--- a/components/AIChatSidePanel.mountRetention.test.ts
+++ b/components/AIChatSidePanel.mountRetention.test.ts
@@ -333,7 +333,7 @@ test('AI side panel skips re-render when only the composer draft text changes', 
   assert.equal(aiChatSidePanelPropsAreEqual(prev, next), true);
 });
 
-test('workspace AI panel memo follows visible inherited session, not hidden terminal maps', () => {
+test('workspace AI panel memo follows visible inherited session, not nonmember terminal maps', () => {
   const visibleMemberChat = session({
     id: 'chat-visible',
     scope: { type: 'terminal', targetId: 'terminal-b', hostIds: ['host-b'] },
@@ -364,7 +364,7 @@ test('workspace AI panel memo follows visible inherited session, not hidden term
       },
     ],
     activeSessionIdMap: {
-      'terminal:terminal-a': 'chat-hidden',
+      'terminal:terminal-outside': 'chat-hidden',
       'terminal:terminal-b': 'chat-visible',
     },
     sessions: [visibleMemberChat, hiddenFocusedChat],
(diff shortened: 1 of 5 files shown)


binaricat:
@codex review

PR #3283, opened 2026-09-06
perf(terminal): skip painting inactive terminal panes

binaricat:
## Summary

Inactive terminal panes can continue repainting while covered by the active tab. Move their screens outside the viewport so xterm can suspend drawing, while continuing to consume output and preserving terminal dimensions. Returning to a tab uses the existing reveal and recovery path.

## Type of Change

- [x] Other (please describe): terminal performance optimization
- [x] Documentation update
- [x] Build / CI change

## Related Issue (optional)

Related to #3278. This targets a verified source of unnecessary background work; it does not establish the cause of the reporter's sustained CPU usage or close the report.

## Changes Made

- Park inactive panes outside the viewport without changing output delivery, buffering, or hibernation policy.
- Keep pinned dimensions valid during window changes, including panes wider than the current viewport.
- Add a real Electron/xterm regression test and run it in Linux CI.
- Document validation and its limits in `docs/research/issue-3278-background-rendering-validation.md`.

## Screenshots / Demo

Validated an isolated full Netcatty instance with five real local shell sessions: continuous background output, repeated tab switches, split panes, focus mode, detaching a pane, and window maximize/restore. Output remained available when returning to a tab.

A local full-app A/B comparison with four continuously writing background shells reduced sampled renderer + GPU CPU usage in all three rounds (approximately 9–45%). This is a short local measurement, not a cross-platform benchmark or a prediction for the reporter's environment. No memory reduction is claimed.

## Testing

- [x] I have tested these changes locally (isolated Vite + Electron development instance)
- [x] Tests pass (`npm test`: 11,377 passed, 18 skipped, 0 failed)
- [x] Production build passes (`npm run build`)
- [x] `npm run test:terminal-background-rendering` passes with real Electron and xterm WebGL
- [x] Regression test fails before the fix: all four inactive terminals continue painting; after the fix they paint zero frames while their buffers still receive output
- [x] Two independent local reviews completed without outstanding findings

Capability generation is not applicable. Existing development warnings were observed; this validation does not claim a warning-free console.

## Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -52,6 +52,9 @@ jobs:
       - name: Test
         run: npm test
 
+      - name: Test terminal background rendering
+        run: xvfb-run -a ./node_modules/.bin/electron --no-sandbox scripts/terminal-background-rendering.live.test.cjs
+
       - name: Test terminal keyword highlight performance
         env:
           NETCATTY_TERMINAL_PERF_SHOW_WINDOW: "1"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -53,7 +53,7 @@ jobs:
         run: npm test
 
       - name: Test terminal background rendering
-        run: xvfb-run -a ./node_modules/.bin/electron --no-sandbox scripts/terminal-background-rendering.live.test.cjs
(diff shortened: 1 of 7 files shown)


binaricat:
@codex review

binaricat:
@codex review

PR #3287, opened 2026-09-06
fix(sftp): settle folder children before history compaction removes their rows

binaricat:
## Summary

A folder could remain active after all files finished because completed child rows were compacted before an old invocation received its superseded response. Large-history recovery could also retain stale failed or paused rows and wait indefinitely. Live and restored transfers now share bounded ownership settlement.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [x] Refactor / code cleanup
- [x] Documentation update
- [ ] Build / CI change
- [ ] Other (please describe):

## Related Issue (optional)

Related to #2568, #3155 and recovery symptoms in #3213; independently reproduced, not original-reporter confirmation.

## Changes Made

- Observe exact terminal identity before history compaction and dispose each observation after its waiter exits.
- Carry current hierarchy identity through recovery events and admit explicit retries without per-file history compaction.
- Distinguish pause waiting, actual cancellation, conflicting identity and completed work. Fresh recovery may resume only unchanged paused rows captured at entry; newer controls still win.

## Screenshots / Demo

Actual live-directory and dedicated-recovery entrypoints reproduce the compacted-child hang. Real child batching at the 4096 threshold covers retained failed/paused rows and a newer pause during source stat. Focused tests include 50k-history batching and observer cleanup.

## Testing

- [ ] I have tested these changes locally (`npm run dev`)
- [x] Linting passes (`npm run lint`)
- [x] Tests pass (`npm test`)
- [ ] Generated capability tool specs are updated when applicable (`npm run generate:capability-tools`)
- [ ] No new console errors or warnings, if this affects app behavior

All five audit branches were combined in an isolated checkout: **11,452 passed, 0 failed, 18 skipped**. Lint and production build passed. Each fix and its follow-ups received independent reviews. These are local results; GitHub checks report their current state separately.

Completed work is never inferred from a missing row or a reused ID with another file identity.

Full evidence and architecture comparison: [SFTP audit report](https://github.com/binaricat/Netcatty/blob/0f7d5fbc91185d489762d47014c24c3d474145c8/docs/research/sftp-transfer-audit-2026-09.md).

## Checklist

- [x] My code follows the existing project style
- [x] I have added or updated relevant documentation
- [x] I have not introduced any breaking changes (or I have described them above)

PR #3291, opened 2026-09-06
fix(sftp): settle direct folder downloads after pause and resume

binaricat:
## Summary

Direct SFTP folder downloads could show Queued while files were transferring. Once paused, resuming could attempt a second directory walk, conflict with the original children, or remain Transferring at 100% after all files reached disk. Use the shared transfer runtime for this entry point and release paused children even after their completed history rows have been compacted.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / code cleanup
- [ ] Documentation update
- [ ] Build / CI change
- [ ] Other

## Related Issue (optional)

Found during smoke testing of changes since v1.1.82, including #3226 and #3287. No separate issue.

## Changes Made

- Register direct downloads as live runtime walks, keep directory status active, and publish final state through the runtime before mirroring it to the panel.
- Keep paused descendant IDs independent of retained UI rows so resume wakes workers whose completed rows were already compacted.
- Derive folder completion counts from the checkpoint and retained children instead of counting a background-completed file again when its invocation settles.
- Add actual hook coverage for pausing discovery with the panel open or unmounted, background completion compaction, exact counts, and no unnecessary reconnect. Add a global-control regression for compacted paused children.

## Screenshots / Demo

Actual `npm run dev` Electron UI on macOS, downloading a 400-file folder from a Linux SSH test host:

1. Download status becomes Transferring and exposes Pause.
2. Pause at 85/400 remains stable; Resume advances normally.
3. Pause again at 367/400, close the SFTP side panel, and resume from the global transfer center.
4. The task reaches Completed, 400 files. All 400 local filenames and file contents match the expected fixture bytes exactly.

The original build reproduced Queued during active transfer. Intermediate repair validation also exposed the lost child pause latch and duplicate count; the final implementation passes the full sequence above.

## Testing

- [x] Tested locally with `npm run dev`, including two pause/resume cycles and panel closure
- [x] `npm run lint` passes
- [x] `npm test` passes: 11,478 passed, 18 skipped, 0 failed
- [x] Generated capability tool specs: not applicable
- [ ] No new console errors or warnings — not asserted beyond the successful observed flows

180 focused transfer tests passed. The first full-suite run encountered 7 unrelated plugin archive failures because this worktree lacked package-local dependencies and resolved yauzl 2.10 instead of 3.4. After restoring the package dependency layout, the full suite passed with 11,478 passed, 18 skipped, and 0 failed. The new direct-download test uses `.test.ts` so the normal test command includes it.

A prior full TypeScript comparison reported the same 732 diagnostics on the baseline and fix, with no new normalized diagnostics at that comparison point; this is not a clean typecheck.

## Checklist

- [x] Existing project architecture and style followed
- [x] Relevant behavior and validation documented above
- [x] No intentional breaking changes

### DIFF
--- a/application/state/sftp/directDownloadRuntime.test.ts
+++ b/application/state/sftp/directDownloadRuntime.test.ts
@@ -1,0 +1,100 @@
+import assert from "node:assert/strict";
+import test from "node:test";
+import React from "react";
+import { act, create, type ReactTestRenderer } from "react-test-renderer";
+import { useSftpTransfers } from "./useSftpTransfers";
+import { transferRuntime } from "./transferRuntime";
+import { sftpTransferCenterStore } from "../sftpTransferCenterStore";
+import { releaseTransferPauseTree } from "./transferPauseLatch";
+
+for (const closePanel of [false, true]) {
+  test(`direct folder download resumes discovery with panel ${closePanel ? "closed" : "open"}`, async () => {
+    const previousWindow = globalThis.window;
+    const previousStorage = globalThis.localStorage;
+    const globals = globalThis as typeof globalThis & { IS_REACT_ACT_ENVIRONMENT?: boolean };
+    const previousAct = globals.IS_REACT_ACT_ENVIRONMENT;
+    globals.IS_REACT_ACT_ENVIRONMENT = true;
+    let finishListing!: () => void;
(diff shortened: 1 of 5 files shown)


binaricat:
@codex review

PR #3275, opened 2026-09-04
Feature/broadcast mode for multiple tabs

Meow-Emperor:
## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / code cleanup
- [ ] Documentation update
- [ ] Build / CI change
- [ ] Other (please describe):

## Related Issue (optional)

N/A

## Testing

- [x] I have tested these changes locally (`npm run dev`)
- [x] Linting passes (`npm run lint`)
- [x] Tests pass (`npm test`)
- [ ] Generated capability tool specs are updated when applicable (`npm run generate:capability-tools`)
- [ ] No new console errors or warnings, if this affects app behavior

## Checklist

- [ ] My code follows the existing project style
- [ ] I have added or updated relevant documentation
- [ ] I have not introduced any breaking changes (or I have described them above)

PR #3292, opened 2026-09-06
fix(terminal): avoid duplicate broadcast keypress text

binaricat:
## Summary

Typing uppercase letters and spaces with cross-tab broadcast enabled could send some characters twice to recipient terminals while the source looked correct. Preserve the physical-key pairing when xterm emits its text on `keypress`, and prevent a trailing `insertText` marker from treating the next physical key as a second composition commit.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / code cleanup
- [ ] Documentation update
- [ ] Build / CI change
- [ ] Other

## Related Issue (optional)

Found while smoke testing #3275 in the development app. No separate issue.

## Changes Made

- Re-arm the existing legacy-data pairing on a matching `keypress`; retain the existing unmatched-event cleanup so later paste is not swallowed.
- Give an identified physical-key payload precedence over a stale composition fallback marker. Actual composition/input handlers clear physical pairing before delivering committed text.
- Add regression coverage that executes the actual registered key, input, and data callbacks with controlled event timing and the real recipient encoder, without constructing the unrelated renderer/addon stack.

## Screenshots / Demo

Verified in the actual `npm run dev` Electron UI on macOS, with local zsh and a Linux SSH session:

- Before: `echo NETCATTY_SMOKE_BROADCAST_20260906_A` looks correct on the source, but recipients repeat uppercase characters. Individual Shift+A/B/C produces `ABBC` at the recipient.
- Captured event order: Space keypress emits data, followed by `insertText`; the next physical A arrives before the composition fallback timer expires and is broadcast again as text.
- After: full uppercase test strings match exactly in both directions; individual Shift+A/B/C produces exactly `ABC`.
- Disabling broadcast keeps a subsequent marker on the source only.
- Pasting `echo "中文 ABC paste"` still reaches the recipient correctly and produces the expected text once.

Temporary diagnostic recording was removed before final verification.

## Testing

- [x] Tested locally with `npm run dev` and Computer UI actions
- [x] `npm run lint` passes; final changed files also pass focused lint
- [x] `npm test`: 11,484 passed, 18 skipped, zero failures
- [x] Generated capability tool specs: not applicable
- [ ] No new console errors or warnings — not asserted for all application flows

118 focused keyboard, broadcast, IME, and terminal tests pass. The baseline fails six delayed-keypress regression cases; the intermediate timing repair still fails the captured trailing-insertText case (` AABBCC` instead of ` ABC`). The final change passes all nine new cases, including keyless Chinese text and subsequent paste.

## Checklist

- [x] Existing project style and input architecture followed
- [x] Behavior and validation documented above
- [x] No intentional breaking changes

### DIFF
--- a/components/terminal/runtime/createXTermRuntime.ts
+++ b/components/terminal/runtime/createXTermRuntime.ts
@@ -1662,6 +1662,13 @@ export const createXTermRuntime = (ctx: CreateXTermRuntimeContext): XTermRuntime
       return false;
     }
 
+    if (e.type === "keypress" && broadcastForwardedKeys.has(kittyKeyIdentity(e))) {
+      // Space and uppercase letters can emit xterm data from keypress, after
+      // keydown's zero-delay cleanup. Rejoin that physical broadcast instead
+      // of sending the same character a second time as unpaired raw text.
+      markBroadcastLegacyDataPending(kittyKeyIdentity(e));
+    }
+
     if (e.type !== "keydown") {
       return true;
     }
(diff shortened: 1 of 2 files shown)


binaricat:
@codex review

PR #3178, opened 2026-08-27
fix(ai): do not report Cursor installed from bundled SDK

binaricat:
## Summary

Settings reported Cursor as installed whenever Netcatty could import its bundled `@cursor/sdk`. That package ships with every build, so the badge was a false positive on machines with no Cursor Agent CLI.

`installed` now means the user has Cursor Agent CLI on PATH or a proven CLI login. `sdkInstalled` still reflects the bundled SDK so API-key mode works without Cursor.app.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / code cleanup
- [ ] Documentation update
- [ ] Build / CI change
- [ ] Other (please describe):

## Related Issue (optional)

Closes #3170

## Changes Made

- Compute Cursor `installed` from `cliBinPath` / `cliLoginOk`, not bundled `@cursor/sdk`.
- Keep `sdkInstalled` and API-key availability (`apiKeyOk && sdkInstalled`) unchanged.
- Settings "Cursor runtime" badge uses CLI presence only; API-key status is unchanged.
- Update resolve-cli / discover tests that encoded the bundled-SDK false positive.

## Screenshots / Demo

N/A — Settings status badge logic only. With no Cursor Agent CLI, the runtime line is "Not detected" even when the bundled SDK is present; API-key mode can still become available after a key is saved.

## Testing

- [x] I have tested these changes locally (`npm run dev`)
- [x] Linting passes (`npm run lint`)
- [x] Tests pass (`npm test`)
- [ ] Generated capability tool specs are updated when applicable (`npm run generate:capability-tools`)
- [x] No new console errors or warnings, if this affects app behavior

Targeted runs:

- `node --test electron/bridges/aiBridge.test.cjs electron/bridges/aiBridge/agentDiscoveryHandlers.test.cjs`
- `node --test --import tsx components/ai/managedAgentState.test.ts components/settings/tabs/ai/types.test.ts`
- `npx eslint` on the touched files

## Checklist

- [x] My code follows the existing project style
- [x] I have added or updated relevant documentation
- [x] I have not introduced any breaking changes (or I have described them above)

PR #3295, opened 2026-09-06
fix(settings): restore AI detection after effect replay

binaricat:
## Summary

In `npm run dev`, opening Settings > AI > Agents can leave every agent stuck on “Detecting…”. React StrictMode replays effect setup and cleanup; the cleanup marks the page unmounted, but setup never restores that flag. Completed CLI checks are then discarded and their loading indicators never clear.

Restore the mounted flag during effect setup while retaining request invalidation on cleanup. This also preserves the existing protection against stale and post-unmount results. Found during release smoke testing; the lifecycle code predates the current release, so this is not attributed to #3178.

## Type of Change

- [x] Bug fix

## Related Issue (optional)

N/A — found during development-app smoke testing.

## Changes Made

- Restore mounted state on effect setup and retain cleanup request invalidation.
- Add tests running the production lifecycle and asynchronous resolver under React StrictMode, with controlled CLI responses. Cover accepted completion, actual unmount, and superseded requests.

## Screenshots / Demo

Verified in the actual Electron application launched with `npm run dev`, using the Settings window. Before: Cursor CLI exists and responds to `--version`, but all cards remain “Detecting…” indefinitely. After a full settings reload with the repair: Cursor shows “Detected” and “Logged in”, the Check button becomes usable, and another manual check completes.

Additional Computer smoke with main cabdcdf31 plus this lifecycle fix used an isolated development profile and a fixture shell environment restricted to system executable directories. The real Cursor resolver could not find `cursor-agent`, while the real bundled SDK import succeeded; diagnostic booleans were `sdkInstalled=true`, `installed=false`, `cliPathPresent=false`, and `cliLoginOk=false`. The current card showed “Not detected / Not logged in”; manual Check completed with the same result. Swapping only the card for its pre-#3178 version produced the original false “Detected” label under the identical backend state, and restoring the current card removed it. No installed executable, user home, credential or normal profile was changed. This validates the CLI-unavailable boundary, not a claim that Cursor was uninstalled from the machine. A fresh 40-test lifecycle/path/auth/discovery subset also passed with no skips.

## Testing

- [x] Tested locally in `npm run dev`
- [x] Focused tests: 63 passed, 0 failed (lifecycle, managed-agent state, path types, discovery and authentication probes)
- [x] Regression tests run against the original implementation: 2 fail; repaired implementation: all 3 pass
- [x] ESLint passes on changed files
- [ ] Full `npm test` was not rerun for this change

## Checklist

- [x] Follows existing project style
- [x] No breaking changes or storage changes
- [x] No capability schema changes

### DIFF
--- a/components/settings/tabs/SettingsAITab.lifecycle.test.ts
+++ b/components/settings/tabs/SettingsAITab.lifecycle.test.ts
@@ -1,0 +1,96 @@
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import test from "node:test";
+import vm from "node:vm";
+import React, { useCallback, useEffect, useRef, useState } from "react";
+import TestRenderer, { act } from "react-test-renderer";
+import ts from "typescript";
+
+(globalThis as typeof globalThis & { IS_REACT_ACT_ENVIRONMENT: boolean }).IS_REACT_ACT_ENVIRONMENT = true;
+
+// Exercise the production lifecycle and asynchronous resolver together, without
+// loading the unrelated provider forms or making real CLI/authentication calls.
+const source = readFileSync(new URL("./SettingsAITab.tsx", import.meta.url), "utf8");
+const lifecycleStart = source.indexOf("  const mountedRef = useRef(true);");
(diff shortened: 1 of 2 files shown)


binaricat:
@codex review

PR #3296, opened 2026-09-06
fix(sftp): preserve browsed paths across tab snapshot transitions

binaricat:
## Summary

Switching away from an SSH terminal and back could still reset its SFTP browser to the terminal cwd and clear its filename filter, despite #3232. The panel briefly consumed the previous tab's live context before the new tab's publisher committed, invalidating cwd tracking and sometimes replacing the SFTP connection.

Keep each panel's last matching terminal/workspace snapshot during that publication gap. Unrelated snapshots are ignored, fresh panels wait for their owner, and actual cwd or focused workspace-pane changes still propagate.

## Type of Change

- [x] Bug fix

## Related Issue (optional)

Follow-up to #3232; related to #3230.

## Changes Made

- Match live snapshots to the owning standalone terminal or workspace.
- Retain only a matching snapshot during a tab reveal; clear the fallback when the hook owner changes.
- Add tests exercising the actual production subscription hook, including stale publication, new mounts, workspace focus changes, and owner changes.

## Screenshots / Demo

Reproduced twice in the actual `npm run dev` Electron app against an SSH host: enable follow terminal directory, browse a different folder, filter to one file, switch to a local tab and back. Before the fix the path returned to `/root` and the filter disappeared. A temporary trace showed the local tab's focused session entering the remote panel before its own snapshot arrived; no panel remount was required.

After the fix, the same UI steps preserved the folder and filter. A real `cd` still moved the SFTP browser to the new directory. Final source was retested in the development app. Diagnostic logging was removed.

## Testing

- [ ] Tests pass (`npm test`) — full suite not rerun
- [ ] Generated capability tool specs are updated when applicable (`npm run generate:capability-tools`) — not applicable
- [ ] No new console errors or warnings, if this affects app behavior — existing development warnings were present

124 focused tests pass across live snapshot isolation, SFTP follow, connection reuse/memory, side-panel wiring, and the tab bridge. The new regression test was run against the old hook and failed because `local-b` reached the `ssh-a` panel; it passes with this fix.

## Checklist

- [x] I have added or updated relevant documentation — reproduction and validation recorded here

### DIFF
--- a/application/state/sidePanelLiveTabIsolation.test.ts
+++ b/application/state/sidePanelLiveTabIsolation.test.ts
@@ -1,0 +1,79 @@
+import assert from 'node:assert/strict';
+import { readFileSync } from 'node:fs';
+import test from 'node:test';
+import vm from 'node:vm';
+import React, { useCallback, useRef, useSyncExternalStore } from 'react';
+import { act, create } from 'react-test-renderer';
+import ts from 'typescript';
+import { getSidePanelLiveSnapshot, subscribeSidePanelLiveSnapshot, sidePanelLiveStore, SIDE_PANEL_INACTIVE_LIVE_SNAPSHOT, type SidePanelLiveSnapshot } from './sidePanelLiveStore';
+
+// Exercise the production hook without importing the unrelated panel UI modules.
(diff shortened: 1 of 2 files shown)


binaricat:
@codex review

Issue #3187, opened 2026-08-28
[Bug] 连接到低端机器上，资源监控导致CPU占用过高


VMatrices:
### Operating system

Linux

### Netcatty version

v1.1.82

### How did you install Netcatty?

GitHub Release (.dmg / .exe / .AppImage / .deb / .rpm / .pacman)

### Affected area

SSH connection / terminal

### Can you reproduce it?

Often (>50%)

### Steps to reproduce

**目标机器环境**
CPU：_Phytium FT-2500 (4 Core)_
系统：_UOS Server 20_

现象是终端连接后发现CPU占用高达98.3%（用户态+系统），排查后发现是监控脚本的问题，kill掉之后占用恢复正常

### Expected behavior

降低对目标机器的系统资源影响

### Actual behavior

在低端机器上占用过高

### Logs / screenshots

```bash
[root@localhost-localdomain ~]# top
top - 10:17:33 up 1 day,  1:40,  5 users,  load average: 4.16, 4.22, 4.37
Tasks: 217 total,   1 running, 216 sleeping,   0 stopped,   0 zombie
%Cpu(s): 47.1 us, 50.6 sy,  0.0 ni,  1.7 id,  0.0 wa,  0.2 hi,  0.3 si,  0.0 st
MiB Mem :  14914.8 total,  11087.6 free,   2628.4 used,   1198.8 buff/cache
MiB Swap:   4095.9 total,   2293.8 free,   1802.2 used.  10470.8 avail Mem

    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
1398919 admin     20   0    0.0t   0.0t   0.0t S   5.9   0.0  41:16.22 /bin/bash -c printf "NC_LATENCY_MARK|"; ostype=$(uname -s 2>/dev/null || echo "Unknown"); if [ "$o+
1394674 admin     20   0    0.0t   0.0t   0.0t S   5.6   0.0  41:18.25 /bin/bash -c exec sh -c 'SELF=$$     TARGET_LOGIN=     ALLOW_HOME_FALLBACK=1     ALLOW_LOGIN_FALLB+
1394682 admin     20   0    0.0t   0.0t   0.0t S   5.6   0.0  40:44.17 /bin/bash -c cat /etc/os-release 2>/dev/null || uname -a
[...]

PR #3301, opened 2026-09-06
fix(ssh): keep remote watchdog alive until blocked children are stopped

binaricat:
## Summary

On Bash-based POSIX shells, a timed-out remote system probe can kill its watchdog before killing the blocked child process. The child remains alive and keeps the output channel open, so repeated probes can still accumulate processes despite #3193.

Replace the command-substitution shell when identifying the watchdog. This returns the watchdog's actual PID, allowing the descendant walk to exclude it and finish killing the blocked probe tree.

## Type of Change

- [x] Bug fix

## Related Issue (optional)

Follow-up to #3193; related to #3187.

## Changes Made

- Add `exec` to the watchdog PID lookup so Bash cannot report an intermediate command-substitution PID.
- Add real-process regressions using the command generated by `getSessionDistroInfo`, with a controlled `cat` executable that either completes or blocks. Assert process-group cleanup, channel closure, and normal successful output without waiting for the watchdog deadline.
- Store test fixtures in Netcatty's managed temporary directory.

## Screenshots / Demo

Reproduced on macOS using the actual generated POSIX probe, without mocking process termination. Before the fix, the five-second watchdog kills the probe shell but leaves the blocked `sleep` alive after seven seconds. After the fix, the channel closes at about five seconds and no live process remains in the test group. A successful probe completes promptly and cleans its watchdog.

An independent process comparison also confirms normal exit status preservation and cleanup of a nested blocked child. The unguarded control remains alive until the test explicitly cleans it up. This validates the shell/process failure itself; the fix has not yet received an additional end-to-end SSH GUI run or a reporter-environment retest.

## Testing

- [x] Tests pass (`npm test`)
- [ ] No new console errors or warnings, if this affects app behavior

Focused SSH stats/CWD/timeout tests: 65 passed. Targeted ESLint passes. Full `npm test`: 11,477 passed, 18 skipped, 0 failed. Full `npm run lint` passed. The initial isolated setup resolved an old root-level ZIP dependency; after restoring the package-local dependency tree, its 29 tests and the full suite passed.

## Checklist

- [x] I have added or updated relevant documentation (inline explanation)

### DIFF
--- a/electron/bridges/sshBridge/sessionOps.cjs
+++ b/electron/bridges/sshBridge/sessionOps.cjs
@@ -91,7 +91,9 @@ function withRemoteWatchdog(command, timeoutMs) {
     `( sleep ${seconds}`,
     // Resolve the watchdog subshell's own PID so the tree walk can skip it
     // (killing ourselves mid-list would abort the remaining kills).
-    `&& nc_self=$(sh -c 'echo $PPID' 2>/dev/null)`,
+    // exec replaces the command-substitution shell: without it Bash may
+    // report that intermediate PID instead of the watchdog subshell.
+    `&& nc_self=$(exec sh -c 'echo $PPID' 2>/dev/null)`,
     `&& nc_tree=$(ps -e -o pid=,ppid= 2>/dev/null | awk -v root="$$" -v self="$nc_self" '${descendantTreeAwk}')`,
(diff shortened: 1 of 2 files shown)


binaricat:
@codex review

PR #3302, opened 2026-09-06
fix(sync): resume auto-sync after first master-key setup

binaricat:
## Summary

First-time encrypted-sync setup in Settings can leave the main window internally locked: it receives the key configuration before the shared password, tries once, and never retries. Host edits then fail to auto-sync until the main window is reloaded. Notify peer windows when the session password becomes available and retry initialization, while preserving deliberate locks and rejecting stale unlock completions.

## Type of Change

- [x] Bug fix

## Related Issue (optional)

Found while smoke-testing #3200 since v1.1.82. This is a pre-existing setup race, not a regression introduced by #3200.

## Changes Made

- Emit a password-availability notification after storing the shared password. The notification contains no password; readers retain the existing protected IPC boundary.
- Retry when the password or key configuration arrives later. Retain the shared password across a failed attempt during key rotation.
- Track unlock/attempt identities per manager across consumers and remounts, preserving intentional locks.
- Reject an unlock result if its key configuration or security generation has changed; preserve an already-installed concurrent winner.

## Screenshots / Demo

Original main development build: initial setup/manual sync succeeds, but host edits produce no automatic upload for over 27 seconds; reloading the main window restores auto-sync.

Latest dab4fe690 development build, driven with Computer Use in a fresh isolated profile and loopback WebDAV service: initial manual upload v1, host creation automatically uploads v2, rename automatically uploads v3. Neither window is reloaded. Settings shows both local and remote v3; the service records all three uploads (8793/9061/9061 bytes). No user cloud account or regular profile was used. The test app and service were stopped, temporary launch settings restored, and normal main development environment restarted.

## Testing

- [x] Latest fresh-profile npm run dev GUI flow above, including the launch lint check.
- [x] Latest full npm test: 11,494 tests, 11,476 passed, 18 skipped, zero failed.
- [x] Latest 15 focused hook, storage, encryption-storage and preload checks pass with no skips.
- [x] Production unlock-method checks with a real derived test key: normal unlock, changed configuration, changed generation, and concurrent winner. Older code fails the changed-configuration control; latest code passes all four. Completion order is controlled in this check rather than forced through the UI.
- [x] Earlier discriminating mounted-hook regressions cover late-password delivery, deliberate lock, remount and replacement-key behavior.
- [x] Generated capability tool specs: not applicable.
- [ ] Latest CI and review completion.

## Checklist

- [x] Existing layer boundaries retained.
- [x] Notification carries no password and does not bypass App Lock.
- [x] Test scope and remaining validation are explicit.

### Pending unlock ownership validation (b9d24020c)

Added two React DOM regressions that keep password reads pending, unmount their first consumer, and then complete all replies. The pre-handoff dab4fe690 implementation fails both normal and StrictMode cases with zero unlocks. The aa50aa811 production change passes both with exactly one unlock; all seven auto-unlock tests and targeted lint pass. This commit adds tests only. Full-suite verification passed: 11,494 tests, 11,476 passed, 18 skipped, zero failures. The late-IPC ownership ordering is controlled by the regression; prior actual first-setup GUI evidence remains at dab4fe690 and is not presented as a forced live ownership race.

### Stale unlock follow-up

Commit e66ece365 addresses review 3944922037. A derivation superseded by a key/configuration change or lock now rejects instead of returning the boolean used for a bad password, including when the old derivation returns no key. This prevents existing sync callers from clearing a newly shared password. A wrong password for the unchanged configuration still returns false.

Three real-crypto stale-completion cases fail before the fix and pass afterward. A rendered `useCloudSync.syncNow` regression exercises the actual caller: the base clears the new password once, the fix never clears it. Thirteen focused checks and full lint pass. Full-suite run is in progress. This follow-up was verified with controlled timing and real derivation; the earlier WebDAV dev-app GUI smoke remains the normal-path evidence. No user vault/password was changed.

### Consumer remount follow-up
Commit 7a677afd9 addresses review 3945305350: a completed empty-password attempt no longer prevents a later consumer from reading a password shared while the page was absent. An active derivation is still deduplicated across unmount/remount. Both normal and StrictMode remount regressions fail on e66ece365 and pass after the fix; an in-flight derivation control remains single. Fifteen focused auto-unlock, real-hook and real-crypto checks pass, with touched-file lint clean. This timing case is covered by React tests, not a live GUI password-sharing claim. Full-suite results above apply to e66ece365, not this new follow-up.

### Shared retry generation follow-up
Commit b8bc61562 addresses review 3944838274. The prior code still included each hook's local retry counter in its shared attempt key. A two-consumer StrictMode regression confirms two derivations on 7a677afd9; after using a manager-scoped password generation, the same test derives once. A notification with two consumers and a later third consumer also shares one attempt. Local state now only wakes the effect and is not part of the attempt identity.

Seventeen focused actual React hook and real-crypto tests pass, zero skipped; touched-file lint and diff checks pass. The full configured suite on b8bc61562 now passes: 11,498 total, 11,480 passed, 18 environment-dependent skips, zero failures (208.4 seconds). The 17 focused React/crypto checks were also run separately. This follow-up is deterministic hook evidence, not a claim that multi-window GUI password rotation has been exercised.

### DIFF
--- a/application/state/useCloudSync.staleUnlock.test.tsx
+++ b/application/state/useCloudSync.staleUnlock.test.tsx
@@ -1,0 +1,65 @@
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import React from 'react';
+import { createDomRenderer, flushEffects, installDomEnvironment } from '../../components/test-support/renderReactDom';
+import { EncryptionService } from '../../infrastructure/services/EncryptionService';
+import type { SyncManagerState } from '../../infrastructure/services/CloudSyncManager';
+import type { SyncPayload } from '../../domain/sync';
+
+test('manual sync does not clear a newly shared password when an older unlock is superseded', async (t) => {
+  const env = installDomEnvironment();
+  const previousStorage = Object.getOwnPropertyDescriptor(globalThis, "localStorage");
+  Object.defineProperty(globalThis, "localStorage", { configurable: true, value: env.window.localStorage });
+  t.after(() => {
+    if (previousStorage) Object.defineProperty(globalThis, "localStorage", previousStorage);
+    else Reflect.deleteProperty(globalThis, "localStorage");
+  });
+  let sharedPassword: string | undefined;
+  let clears = 0;
+  Object.defineProperty(env.window, 'netcatty', { configurable: true, value: {
+    cloudSyncGetSessionPassword: async () => sharedPassword,
+    cloudSyncClearSessionPassword: async () => { clears += 1; sharedPassword = undefined; },
+  } });
+  const { getCloudSyncManager } = await import('../../infrastructure/services/CloudSyncManager');
+  const manager = getCloudSyncManager();
+  await manager.setupMasterKey('old-fixture-password');
+  manager.lock();
+  const newConfig = await EncryptionService.createMasterKeyConfig('new-fixture-password');
+  const { useCloudSync } = await import('./useCloudSync');
+  let current!: ReturnType<typeof useCloudSync>;
+  const Probe = () => { current = useCloudSync(); return null; };
+  const renderer = await createDomRenderer(env.document);
+  const originalUnlock = EncryptionService.unlockMasterKey;
+  let release!: () => void;
+  let started!: () => void;
+  const gate = new Promise<void>((resolve) => { release = resolve; });
+  const entered = new Promise<void>((resolve) => { started = resolve; });
+  try {
+    await renderer.render(<Probe />);
+    await flushEffects();
+    EncryptionService.unlockMasterKey = async (...args) => {
+      started();
+      await gate;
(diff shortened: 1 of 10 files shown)


binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

PR #3297, opened 2026-09-06
fix(terminal): preserve row boundaries in alternate history preview

binaricat:
## Summary

Shift+PageUp inside Vim could concatenate separate lines in the alternate-screen history preview because the collector discarded cursor-addressed row transitions. Preserve those transitions and terminal cell positions so separate rows, empty lines and wide-character spacing remain readable. Resizing now also invalidates cached preview text and row counts when the open line is trimmed, preventing stale characters from reappearing after the terminal returns to its previous width.

## Type of Change

- [x] Bug fix

## Related Issue (optional)

Follow-up to #3165, related to #2516. Found during post-v1.1.82 development-app smoke testing.

## Changes Made

- Track supported row/column controls, scrolling margins, saved cursor state and automatic wrapping across output chunks and viewport changes; reset retained state between terminal boots.
- Preserve soft-wrap metadata, explicit empty rows and whole graphemes at overwrite/erase boundaries while keeping bounded transcript limits.
- Invalidate the preview cache when narrowing truncates the open line, even if no preview is read at the intermediate width.
- Synchronize history from the actual terminal resize event, including its cursor position and visible cursor row. This handles idle applications, normal-buffer reflow and alternate-buffer retained cells without guessing screen positions from transcript history.
- Preserve a final wide glyph under no-autowrap overflow, and restore DECAWM/DECOM with saved cursor coordinates. Tests compare both against the bundled real xterm.
- Cover the captured Vim case, Unicode and control boundaries, and cached text/count consistency after resizing.

The preview remains a bounded text transcript; it does not reconstruct every full-screen layout or scrolling-region operation. In particular, revisiting a previously committed screen row records a new transcript entry rather than retroactively replacing the earlier entry. Full screen-row reconstruction requested by the latest review is outside this PR: it changes the history-retention model, so this behavior remains an explicit limitation.

## Screenshots / Demo

[...]

PR #3304, opened 2026-09-06
fix(terminal): keep full-width history preview rows visible

binaricat:
## Summary

Full-width terminal output is readable in xterm, but the native history preview clips the final columns and packs rows more tightly. The preview uses the same font options but misses xterm's device-pixel-rounded cell dimensions. Fit each preview row to its live terminal cell width and match the rendered line height.

## Type of Change

- [x] Bug fix

## Related Issue (optional)

Found during the post-v1.1.82 Computer smoke of #3165. Independent of the transcript parser repairs in #3297.

## Changes Made

- Scale each native preview row to its terminal cell width, including mixed ASCII/CJK rows, instead of applying a per-glyph letter-spacing correction.
- Batch row measurements before transforms and match line height and font weight to the terminal.
- Keep explicit newline text nodes and calculate DOM selection offsets across row spans, preserving hard/soft-wrap copy behavior.
- Select the entire overlay for keyboard/context-menu Select All so every row is included.
- Retain the plain-text fallback when terminal dimensions are unavailable. Incoming terminal output processing is unchanged.

## Screenshots / Demo

Actual `npm run dev` Electron, operated with Computer on macOS. A Python fixture emits 100 full-width ASCII lines ending in END in the alternate screen. Before: END visible in the live terminal, clipped from every preview row. After: END visible in ASCII preview at both sidebar-open and sidebar-collapsed widths. The wide-glyph follow-up also shows END on full CJK rows; actual Computer copy into TextEdit retains CJK-000, the Chinese characters, and END. ASCII was rechecked after that follow-up; exit returns to the shell. The Select All follow-up was validated through actual Cmd+A/Cmd+C and paste into TextEdit: all 51 displayed CJK rows (000 through 050), each with 93 Chinese characters and END, were retained. A new regression fails before the fix by selecting only the first row and passes after it. Local artifacts: preview-edge-main-live.png, preview-edge-main-preview.png, preview-edge-fixed.png, preview-edge-fixed-wide.png in the smoke evidence folder.

## Testing

- [x] Linting passes on the changed file
- [x] Related tests pass: 44 createXTermRuntime / terminalHistoryScrollOverride tests, including multi-row Select All, nested-row partial and soft-wrap selection
- [x] Latest commit 39d4fcfee CI lint-and-test passed (run 34056098689 / job 101548165752)

Validation is on the configured macOS font and actual app window, not every font, platform, or emoji/grapheme combination. The added measurement runs when opening/paging history, not on incoming terminal output; Measured isolated Electron render-block A/B (40 warmed alternating samples) adds about 0.7ms per page for ASCII/CJK at 196 columns x 52 rows; emoji adds 1.3ms (6.2 to 7.5ms). At 512 x 100, dense emoji increases from 32 to 34.9ms, so this is not a zero-cost change. The benchmark excludes history parsing/input dispatch/compositor presentation. Actual Computer paging through a 300-row ASCII/CJK/emoji fixture moved backward and forward to the expected row labels with END visible and exited back to the shell.

## Checklist

- [x] Follows the existing runtime structure
- [x] No stored-data or public API change

### DIFF
--- a/components/terminal/runtime/createXTermRuntime.ts
+++ b/components/terminal/runtime/createXTermRuntime.ts
@@ -1065,6 +1065,27 @@ export const createXTermRuntime = (ctx: CreateXTermRuntimeContext): XTermRuntime
     overlay.style.fontSize = `${currentTerminalFontSize()}px`;
     overlay.style.fontFamily = String(term.options.fontFamily ?? fontFamily);
     overlay.style.lineHeight = String(term.options.lineHeight ?? lineHeight);
+    // xterm rounds its cell dimensions to device pixels. A native <pre> with
+    // the same font settings can therefore be wider (and shorter) than the
+    // terminal, clipping the final columns of every full-width history row.
+    const screen = term.element?.querySelector(".xterm-screen");
+    const screenRect = screen?.getBoundingClientRect();
+    if (screenRect && screenRect.width > 0 && screenRect.height > 0
+      && term.cols > 0 && term.rows > 0) {
+      overlay.style.fontWeight = String(term.options.fontWeight ?? "normal");
+      overlay.style.fontKerning = "none";
+      overlay.style.fontVariantLigatures = "none";
+      overlay.style.letterSpacing = "0px";
+      const measure = document.createElement("span");
+      measure.style.display = "inline-block";
+      measure.textContent = "W";
+      overlay.replaceChildren(measure);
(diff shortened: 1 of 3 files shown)


binaricat:
@codex review

binaricat:
@codex review

binaricat:
@codex review

PR #3239, opened 2026-09-01
fix: make terminal drag uploads route-safe

binaricat:
## Summary

Terminal file drops after sudo/su no longer guess a login-shell or home directory. A drop uses a confirmed current directory and the originating terminal's SFTP session. Requested sudo SFTP no longer silently becomes ordinary SFTP/SCP. Slow jump-host ZMODEM starts keep the longer wait window.

This does not auto-enable sudo SFTP so unelevated SFTP can write `/root`. If the active shell directory cannot be confirmed, the drop fails with a destination-unknown error instead of uploading to the wrong place.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [x] Refactor / code cleanup
- [ ] Documentation update
- [ ] Build / CI change
- [ ] Other (please describe):

## Related Issue (optional)

Related to #3237

## Changes Made

- Stop guessing upload directories after shell identity or working-directory changes
- Bind SSH drops to the originating terminal session
- Keep requested sudo SFTP from falling back to ordinary SFTP/SCP
- Keep slow ZMODEM/SFTP starts from being abandoned at the old 2.5s window
- Remove extra queued-drop routing (dormant/activated queue and a third route-ownership id)

## Screenshots / Demo

Not applicable; this changes connection and upload behavior without visual changes.

## Testing

- [ ] I have tested these changes locally (`npm run dev`)
- [x] Linting passes (`npm run lint`)
- [x] Tests pass (`npm test`)
- [ ] Generated capability tool specs are updated when applicable (`npm run generate:capability-tools`)
- [ ] No new console errors or warnings, if this affects app behavior

Additional verification:

- Focused terminal/SFTP unit tests of the shipped cwd, drop, and sudo-open helpers
- Electron live sudo-i / jump-host drag still needs a human reproduction before closing #3237

## Checklist

- [x] My code follows the existing project style
- [ ] I have added or updated relevant documentation
- [x] I have not introduced any breaking changes (or I have described them above)

PR #3432, opened 2026-09-18
fix(ssh): reject lsof error text as terminal upload paths

binaricat:
## Summary

After `sudo -i`, an unreadable current directory can make lsof return `/proc/<pid>/cwd (readlink: Permission denied)` as its name field. Netcatty accepted this diagnostic as a real directory and used it as an upload destination, matching the screenshot in #3237. Accept only lsof records identified as directories so the existing destination-unknown handling applies instead.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / code cleanup
- [ ] Documentation update
- [ ] Build / CI change

## Related Issue (optional)

Related to #3237. This fixes the diagnostic-as-directory case left after #3239 and #3241; it does not establish that every sudo/jump-host/reconnect scenario in the issue is resolved.

## Changes Made

- Request lsof file types and use only directory records in the cwd fallback.
- Exercise the actual generated shell helper with permission/missing-path diagnostics, absent type information, normal paths, Unicode escapes, and a legitimate directory whose name contains diagnostic-looking text.

## Screenshots / Demo

No UI change. Actual Debian Linux reproduction using the generated remote command:

- Before: unprivileged access to a root-owned process emits `NETCATTY_LSOF_CWD=/proc/1/cwd (readlink: Permission denied)`.
- After: the same probe exits 1 with no directory; a privileged probe still returns `/`.
- On macOS, the real lsof fallback returns this worktree's current directory.

## Testing

- [x] Linting passes (changed files)
- [x] Tests pass (focused cwd, sudo, elevation, drop, and reconnect suites)
- [ ] No new console errors or warnings, if this affects app behavior

The new regression fails before the fix and passes afterward. Two independent local reviewers returned CLEAN. No full suite or Electron end-to-end claim; the reported jump-host configuration is unavailable.

## Checklist

- [x] I have added or updated relevant documentation (inline explanation and reproduction here)


binaricat:
@codex review

